PR #84, opened 2015-04-12
Generic AppiumDriver

TikhomirovSergey:
Hi guys! I want to propose changes that are similar as https://github.com/appium/java-client/pull/182 (java client, it is still waiting for the merging or rejection).

Major change:
AppiumDriver and its subclasses are generic now. This feature will allow to user the choice of IWebElement type those instances should be found and returned.

```c#
AppiumDriver<IWebElement> driver;
//some actions

IWebElement e = driver.FindElement(By.Id("id"));
List<IWebElement> l = driver.FindElements(By.id("id"));
```

User can choose the desired type. IWebElement, RemoteWebElement, AppiumWebElement or subclasses are allowed. So when they will choose something that differs from IWebElement then:

```c#
AppiumDriver<AppiumWebElement> driver;
//some actions

AppiumWebElement e = driver.FindElement(By.Id("id"));
IList<AppiumWebElement> l = driver.FindElements(By.Id("id"));

AppiumWebElement e1. = e.FindElement(By.Id("id2"));
IList<AppiumWebElement> l1 = e.findElements(By.Id("id2"));
```
AppiumWebElement and subclasses are not generic.  But their finders already return AppiumWebElement.

@Astro03 @Jonahss @bootstraponline What do you think about changes like that in C#?
I would like to merge them but release them later. There is scope of changes that I am going to implement and release as 2.0.0.0 version.

### DIFF
--- a/appium-dotnet-driver/Appium/Android/AndroidDriver.cs
+++ b/appium-dotnet-driver/Appium/Android/AndroidDriver.cs
@@ -6,15 +6,16 @@ using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics.Contracts;
-using System.Linq;
-using System.Text;
 
 namespace OpenQA.Selenium.Appium.Android
 {
-    public class AndroidDriver : AppiumDriver, IFindByAndroidUIAutomator, IStartsActivity, IHasNetworkConnection, 
-        IAndroidDeviceActionShortcuts, IPushesFiles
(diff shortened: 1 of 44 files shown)


Astro03:
I'm okay with doing it that way if that's how the java guys do it.. I agree
that we need to be in line with how they do things as well.

Astro03:
sorry.. didn't see this until now..

Issue #85, opened 2015-05-16
Should be able to set GPS location


poindextrose:
The JSON wire protocol supports setting the location. This driver have this command.

poindextrose:
This works on iOS but on Android I get
```
OpenQA.Selenium.WebDriverException : Unexpected error. ERROR running Appium command: port should be > 0 and < 65536: 0
```

PR #86, opened 2015-05-16
set location of device

poindextrose:
issue #85

### DIFF
--- a/appium-dotnet-driver/Appium/AppiumDriver.cs
+++ b/appium-dotnet-driver/Appium/AppiumDriver.cs
@@ -325,6 +325,20 @@ namespace OpenQA.Selenium.Appium
             this.HideKeyboard(null, null);
         }
 
+        /// <sumary>
+        /// GPS Location
+        /// </summary>
+        public Location Location { 
+                get {
+                        var commandResponse = this.Execute(AppiumDriverCommand.GetLocation, null);
+                        return JsonConvert.DeserializeObject<Location>((String) commandResponse.Value);
+                }
+                set {
+                        var location = value ?? new Location();
+                        Execute(AppiumDriverCommand.SetLocation, location.ToDictionary());
+                }
+        }
+
         #endregion MJsonMethod Members
 
         #region Context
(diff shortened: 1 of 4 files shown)


TikhomirovSergey:
+1 I like it.
But... Can you provide some tests? It would be cool if there were one test on Android and one on iOS.

poindextrose:
I would love to write some tests, but I'm having trouble getting the existing tests to work. My environment is OS X with Xamarin Studio. I get the following when I run the `MJsonMethodTest`
```
WebDriverException: Unexpected Error.
```
The only test I can get to pass from the `test` project is `InitSessionTest`

Does anyone have any tips on getting the tests to work?

TikhomirovSergey:
@poindextrose 
Do you mean these tests? [appium-dotnet-driver/test/specs](https://github.com/appium/appium-dotnet-driver/tree/f9ae7a7c55dbc65ba7a2e2801a595fc7c927bcb5/test)

Actually I don't know what to do with them :( This is sad. Maybe @Astro03 knows more. Maybe they are outdated. Can you exclude this project on your environment?

Actually I use [samples](https://github.com/appium/appium-dotnet-driver/tree/master/samples) for the testing now (as a temporary solution). You can add the sample here too.

poindextrose:
Yeah that is what I was talking about. I'll look into the samples

TikhomirovSergey:
Also I have one more remark. Actually it is possible to get location as well as set it. 
Please look at these:
[Location](https://github.com/SeleniumHQ/selenium/blob/6b4e6de722fcccd18619951543143999e9a30657/java/client/src/org/openqa/selenium/html5/Location.java) 
[RemoteLocationContext](https://github.com/SeleniumHQ/selenium/blob/6b4e6de722fcccd18619951543143999e9a30657/java/client/src/org/openqa/selenium/remote/html5/RemoteLocationContext.java)

What is there were the __Location__ class that contained GPS coordinates and the property that could set/get these coordinates?

Astro03:
Honestly I have not touched any of the tests so I wouldn't know about it with Xamarin.   

@sebv was the one that wrote all those I believe.  

Ciao,
Astro
Sent from my phone

On May 16, 2015, at 15:55, Sergey Tikhomirov <notifications@github.com> wrote:

@poindextrose 
Do you mean these tests? appium-dotnet-driver/test/specs

Actually I don't know what to do with them :( This is sad. Maybe @Astro03 knows more. Maybe they are outdated. Can you exclude this project on your environment?

Actually I use samples for the testing now (as a temporary solution). You can add the sample here too.

poindextrose:
Good idea. I'll look into that.

Also I getting a warning in the Appium console output when I use this new `SetLocation`:
```
warn: Use of the set location method with the latitude and longitude params as top-level JSON params is deprecated and will be removed. Please update your client library to use a method that conforms to the spec
```

I'm very new to Appium, so I'm having trouble finding the spec that it is referring to.

TikhomirovSergey:
@poindextrose 
This is the warning from the Appium server: https://github.com/appium/appium/blob/0ca209bcbc3d5fe48409190f53ba28b41584bde5/lib/server/controller.js#L854
You have to define all coordinates. Now you are trying to provide the ability to use them optionally:
https://github.com/poindextrose/appium-dotnet-driver/blob/set-location/appium-dotnet-driver/Appium/AppiumDriver.cs#L331
This is wrong. So the Location-class looks as more suitable solution.

poindextrose:
I've updated the code to use a property along with the ability to get the location. However the Appium server does not implement the get location method, so if you try to get the location, a not implemented exception will be thrown from the server, see [routing.js](https://github.com/appium/appium/blob/84e84584042d0242704c8a98e92b87a7ac74ceda/lib/server/routing.js#L166)

TikhomirovSergey:
Ok! Today I'll review and merge this.

TikhomirovSergey:
Ok! I have checked it. Tomorrow I'm going to add a test/sample.
...About the getting of the location. It's really not implemented command. I'm feeling the upset. :( The same problem is on the java_client side too. I'm going to report a new issue [here](https://github.com/appium/appium/issues). @poindextrose If want you can report it too.
Thank you.

PR #95, opened 2015-06-19
add xml doc tags

Astro03:


Astro03:
also add "new" keyword to most methods in appiumdriver since it overrides base class's methods.

PR #96, opened 2015-07-06
Some features ported from the Java-Appium-Driver

trbnb:
AppiumDriver
- Tap
- Swipe
- Pinch
- Zoom
- ScrollTo
- ScrollToExact

IOSDriver
- GetNamedTextField

AndroidDriver
- GetNetworkConnection
- SetNetworkConnection
- also ScrollTo & ScrollToExact implementations with extra parameter for a resource ID to scroll on a particular View.


Astro03:
1) You're missing your csproj file which adds the recently new files into the project. You're solution will not compile.
2) missing xml doc tags on all newly created methods
3) Change namespace of IScrollsTo.cs to "OpenQA.Selenium.Appium.Interfaces" - this will match the other interfaces.
4) AppiumDriver - write new IScrollsTo Methods in a region
5) AndroidDriver.NotNullOrEmpty(), a more elegant solution would be "return !string.IsNullOrEmpty(thing as string);"  (I'm assuming this is what you are trying to do.)
6) AndroidDriver.SetNetworkConnection - Just curious why you call the GetCommandImmutableMap() instead of just doing a dictionary.Add(string, object) - I'm only worried because the GetComamndImmutableMap does not check for length of values and you can get either a NullException or IndexOutOfRangeException.

just looking from the top of my head  - if you can please fix those. 

Thanks,
ASTRO

TikhomirovSergey:
@Astro03  @trbnb  Hi. I've superficially reviewed it and I've got some remarks.

>AndroidDriver
•GetNetworkConnection
•SetNetworkConnection

It duplicates the existing feature. Please look at this:
https://github.com/appium/appium-dotnet-driver/blob/master/appium-dotnet-driver/Appium/Android/AndroidDriver.cs#L103

trbnb:
>It duplicates the existing feature. Please look at these:

>https://github.com/appium/appium-dotnet-driver/blob/master/appium-dotnet-driver/Appium/Android/AndroidDriver.cs#L103

>https://github.com/appium/appium-dotnet-driver/blob/master/appium-dotnet-driver/Appium/Android/Enums/ConnectionType.cs

OK, I must have missed this because of the different naming. 

I do wonder though why there is an extension method to convert `object`s into `ConnectionType`s. I fail to see how this makes sense. Could you explain?

TikhomirovSergey:
@trbnb

If I correctly understood you

There is the enum on .Net client side. [ConnectionType](https://github.com/appium/appium-dotnet-driver/blob/master/appium-dotnet-driver/Appium/Android/Enums/ConnectionType.cs). Each object has a number value that equals to connection bit map.  

So we don't have to convert enum object to a number (it is already number). But we have conver the number to some enum object. 

Please read about C# enums:  http://csharp.net-informations.com/statements/enum.htm

trbnb:
I know how enums work. I just cant figure out why you used an extension method to convert the value of the response into your ConnectionType enum. Makes no sense to me.

TikhomirovSergey:
Ok. Actually I didn't implement this (just refactored the solution structure). Maybe @Astro03 could explain it.

If you consider that it is the excessive action you are free to improve it. And you can check this change via this test/sample: https://github.com/appium/appium-dotnet-driver/blob/master/samples/AndroidConnectionTest.cs

Astro03:
When coding, I would rather work with concise list of objects (enums)
rather than have random numbers going around.

With changes you propose, many integer values given into the constructor
would make the network connection be a WiFi (such as 10). The proposed
method forces people to do math to figure out which bits are set and which
are not -> allowing a possible source of bugs within their code.

In the original code, we are removing a layer discrepancy and forcing
people to know/choose exactly what kind of network setting they would like.

Does this make sense? Can we make improvements to this? Is your proposed
solution better?

Thanks,
ASTRO

trbnb:
So, apparently I wasn't really making it clear enough. I have no critics in terms of using an enum in this situation. (Even though it is a bit weird that the Java and C# implementations have this difference.)

What I don't understand is why you use an extension method to convert your response value. Why would you want to make this method a member of every single instance of anything if I only need it in one specific situation?

Astro03:
Ahh, understood.

I honestly don't remember why I used an extensions method here but I just
thought it would be a cool place to use it. AND the fact that I liked
extension methods.

I guess another way would have been to have a static class which converted
it for us. This should probably have been the way we did it instead since
you are correct, every object now by definition has this extension method
associated with it.

TikhomirovSergey:
@trbnb 
I found the last change ok. :+1: 
Could you provide some tests/samples?

Astro03:
@trbnb - hey there, i'm about to pull in @TikhomirovSergey 's changes and will publish a new dotnet driver. if you want these changes to be pushed in, please implement the code review comments.. thanks man

trbnb:
OK, I will try and have a look into it tomorrow morning.

trbnb:
Hey, I added a test for the gestures that works on Android. I didn't really have time for much more so that's all I can do for now. I hope your happy with it.

TikhomirovSergey:
@trbnb 
:+1:

Issue #107, opened 2015-09-15
Apache-2.0 Licensing


TikhomirovSergey:
.Net client should follow the apache 2.0 license. It means:

- [having a license file in the repo](https://github.com/SeleniumHQ/selenium/blob/master/LICENSE)
- [a notice file with contributors](https://github.com/SeleniumHQ/selenium/blob/master/NOTICE)
- [an appropriate header file on each source file](https://github.com/SeleniumHQ/selenium/blob/master/dotnet/src/webdriver/Remote/RemoteWebDriver.cs)

PR #111, opened 2015-10-11
#107 fix: Licensing

TikhomirovSergey:
Change list:
- all unused imports were removed from sources

- Appium .Net client is following the Apache 2.0 license.


Astro03:
looks good to me

PR #120, opened 2015-11-14
#108 fix

TikhomirovSergey:
Change list:
- redudant constructors 
 
```c#
public TouchAction ()

public MultiAction(IWebElement element)

public MultiAction()
```

were removed because they don't populate necessary elements and then look useless. Also they cause problems #108 

Now this upstream should be more synchronized with Java client.

- improved code style


TikhomirovSergey:
@Astro03 
Can you take a look at this PR?

Astro03:
awesome. I'll check it out in a day or two when i'm in front of a computer
:)

On Sat, Nov 14, 2015 at 1:51 PM, Sergey Tikhomirov <notifications@github.com

Astro03:
looks awesome. this should relieve some confusion :)

PR #122, opened 2015-11-17
Patch to enable xmldoc

pver:
This patch enables xmldoc generation:
- during a release build, xmldoc file is generated
- during creation of the nuget package, the xmldoc is included in the nuget package

This enables to see the xml documentation using intellisense:
image


Jonahss:
Cool Thanks!
@Astro03  look good?

Astro03:
i believe @TikhomirovSergey  is going to be publishing a new one today

PR #132, opened 2016-01-15
New Zoom/Pinch implementation

trbnb:
After working with the zoom and pinch functionalities I made in my last PR we (myself and a colleague) came across some inconsistencies. 

We figured that iOS and Android use very different approaches to these gestures, so we worked on a new implementation for these that works much more consistent.


Astro03:
do you want to take out the generic approach all together?

trbnb:
Well, the problem is that the Appium server needs different parameters for the different platforms. So the only alternative to this approach would be to define an interface with **all** of those parameters. But this way one implementation would just ignore a half of these parameters and the *other* implementation would ignore the *other* half. And I think this doesn't really add anything useful to the code.

Astro03:
Was there something wrong with the multi touch way of doing things instead
of running the script directly?

trbnb:
Yes, after observing it a bit we found it to be quite unreliable. Especially on iOS we sometimes struggled to make it work.

TikhomirovSergey:
I'll review this. I've some remark...

TikhomirovSergey:
Sorry for the late responce, I was busy

I have a a question to @Jonahss. Let him to take a look at your implementation. I have some question to the original code... If it is correct then it makes sense to port your code to Java client.

trbnb:
I'm not really sure how because the gestures don't change the size the view or something like this, just how the content is presented and I don't think that's something we can retrieve with Appium.

trbnb:
Any news on this?

TikhomirovSergey:
Waiting for @Jonahss (I've told about this PR) because I think that this change could be ported to Java client too. 

As I know current gestures were ported from Java client. I dont know why original code was designed the way as it is now. But it is most likely I'm Ok with your changes.

trbnb:
Ok, good to know. 

And yes, these gestures were initially ported from the Java client, I made the pull request for that a few weeks/months ago.

TikhomirovSergey:
@trbnb 
i'm ok with your changes. 
There are tons of work with appium node 1.5.x. I'm working on java client now. You fix is supposed to be included in 1.6.0.1. There will be many fixes related to node 1.5.x.

trbnb:
Alright, thanks for letting me know.

BaN4NaJ0e:
Just for your information. We (trbnb and myself) ported the new pinch and zoom behavior from the python appium driver. More information can be found here: 

https://github.com/appium/python-client/blob/master/appium/webdriver/webdriver.py

Be careful. They are calling pinch and zoom in the code a little bit different. It is internally called pinchClose and pinchOpen. But it does the same.

It is a good idea to port our PR to the java bindings too. Pinch and Zoom is broken there too. We initialy ported the pinch and zoom command from the java bindings to the dotnet bindings without testing if it worked in the Java bindings. Seems that it did not work under iOS at all. So thats the reason we did it again. This time we ported it from the python bindings. And now it should work on Android and iOS.

TikhomirovSergey:
Guys,
I'm going to play around with proposed changes. I think I'll propose a PR with your authorship if everything is ok.

However, I'm not sure that it is compatible with node server 1.5.x

TikhomirovSergey:
Sorry.
I need to close it without merging. This functionality is not supported since Appium 1.5.0
https://github.com/appium/appium/blob/master/CHANGELOG.md#general-3

Server is throwing this exception now.
>Method has not yet been implemented (WARNING: The server did not provide any stacktrace information)

See also:
https://github.com/appium/appium/issues/2234

Jonahss:
Can we add docs for the new way of performing pinch and zoom?

TikhomirovSergey:
@Jonahss It has to work the same way.
I and @ShrivanTarget are searching for something instead.

Issue #134, opened 2016-01-20
Implement `/wd/hub/session/:sessionId/appium/device/system_time` endpoint


imurchie:
https://github.com/appium/node-mobile-json-wire-protocol/issues/40

TikhomirovSergey:
Ok.

TikhomirovSergey:
Fixed at #141

Issue #136, opened 2016-01-31
Appium 1.5 support


TikhomirovSergey:
Related client and server bugs will be mentioned here

TikhomirovSergey:
https://github.com/appium/appium/issues/5936

PR #141, opened 2016-02-19
#134 FIX

TikhomirovSergey:
Change List:
- #134 FIX. The DeviceTime property was added and it works with Appium
node 1.5

-  improvements of locking methods. The LockDevice(seconds) is obsolete
and it is going to be removed in the next release. Since Appium node
server v1.5.x it is recommended to use
AndroidDriver.Lock()()...AndroidDriver.Unlock() or IOSDriver.Lock(int
seconds) instead.

### DIFF
--- a/appium-dotnet-driver/Appium/AppiumDriver.cs
+++ b/appium-dotnet-driver/Appium/AppiumDriver.cs
@@ -902,7 +902,10 @@ namespace OpenQA.Selenium.Appium
         #endregion
 
         #region Device Time
-
+        /// <summary>
+        /// Gets device date and time for both iOS(Supports only real device) and Android devices
+        /// </summary>
+        /// <returns>A string which consists of date and time</returns>
         public String DeviceTime
         {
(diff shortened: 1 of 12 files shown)


TikhomirovSergey:
@Astro03 @imurchie Could you take a look at this PR.

TikhomirovSergey:
#136

imurchie:
This looks reasonable to me, without any knowledge of the language.

appium-dotnet-driver/Appium/AppiumDriver.cs
```diff
@@ -898,6 +901,19 @@ public void Zoom(IWebElement el)
 
         #endregion
 
+        #region Device Time
+
```
Astro03: add doc tags here.. would be good for people using this class to know what this property is used for.

Astro03:
looks silky otherwise. i like how you put the obsolete tag on there :)

Issue #168, opened 2017-02-01
It is no longer possible to override the Execute method.


Siko91:
## Description

Until recently, I was using an older version of Appium for iOS automation.
Then I updated the devices, mac, XCode, and appium server and client.

When updating the client, I noticed that it is no longer possible to ovverride the "Execute" method from AppiumDriver, because it was declared with the keyword 'new'.

```
lines 305-306 in AppiumDriver.cs

public new Response Execute(string commandName, Dictionary<string, object> parameters) => 
            base.Execute(commandName, parameters);
```

It is pretty obvious that stopping people from overriding this method was intentional.
The keyword 'new' was put, in order to hide the base method from the inheriting classes.
Why was this choice made?

I am using this method a lot.
Here's the override code that I use: [Appium Execute method override](http://pastebin.com/1P0T9PX7)

As you can see, I use it for:

- Collecting information about every interaction with the server.
- Collecting information about every searched element.
- Logging (a lot). The logging happens in my collect methods.
- Providing more verbose exception messages. (The exceptions from the appium server didn't really state which action failed - only that there is a failure)
- Deciding (based on the type of error) when to mark the test as inconclusive, and when to mark it as failed

TikhomirovSergey:
Ok. I will try to improve that ASAP and publish a new build. I was sure that nobody was using that.

TikhomirovSergey:
Was published at 3.0.0.2

PR #169, opened 2017-02-03
Explicit interface impentation, not to hide RemoteWebDriver Execute method

sttodorov:
With the explicit interface implementation is possible to define both methods form RemoteWebDriver and from IExecuteMethod.
This can be applied also to all other methods defined with "new" in AppiumDriver like FindElements, FindElementById, etc. 

Related to [issue 168](https://github.com/appium/appium-dotnet-driver/issues/168)
 
- [X ] Bugfix (non-breaking change which fixes an issue)


TikhomirovSergey:
@sttodorov Thank you very much

PR #208, opened 2018-01-12
4.0.0.BETA1. Build number.

TikhomirovSergey:
## Change list

Build number was changed to 4.0.0.BETA1. 
 

## Details

The starting of 4.x seria.


TikhomirovSergey:
@Astro03 Could you have a look? Is it ok?

PR #216, opened 2018-04-06
Add element attribute caching

mgkeeley:
## Change list

Add element attribute caching to AppiumWebElement. This builds on top of the work done in
https://github.com/appium/WebDriverAgent/pull/69
https://github.com/appium/appium-uiautomator2-server/pull/128
https://github.com/SeleniumHQ/selenium/pull/5731

Caching is opt-in.

Note that his change is dependent on the changes in https://github.com/SeleniumHQ/selenium/pull/5731, it will not compile until these changes have been promoted to nuget and the references to WebDriver in this project have been updated.

## Types of changes

## Details
By setting the capability `"shouldUseCompactResponses": false`, extra attributes are returned from find-element calls. The list of attributes is specified with the capability `elementResponseAttributes`, using W3C names.
With this feature, if the `shouldUseCompactResponses` capability is *false*, then the cache is enabled for the elements. Returned attributes are stored in the AppiumWebElement, and getting the corresponding proeprties on AppiumWebElement will return the values from the cache. Attributes not in the cache will be fetched from the target and added to the cache.
A new interface IWebElementCached allows some control over the cache.


appium-dotnet-driver/Appium/AppiumWebElement.cs
```diff
@@ -48,11 +50,111 @@ public abstract class AppiumWebElement : RemoteWebElement,
...
         {
+        }
+
+        #region Cache 
+
+        protected Dictionary<string, object> cache = null;
+
+        public virtual void SetCacheValues(Dictionary<string, object> cacheValues)
+        {
+            cache = new Dictionary<string, object>(cacheValues);
```
mykola-mokhnach: I'd rather use more advanced data structure for the cache, for example LRU cache with limited items size, since the "classic" dictionary may grow up too big if there are many elements in the UI.

mgkeeley: The cache is per element; so there will only be a handful of attributes in this dictionary

mgkeeley:
@mykola-mokhnach This PR is now ready to be reviewed and merged.
I have updated the dotnet client to use Selenium 3.13.0, and implemented the new ElementFactory method for element creation.
Cheers, Mike.

mykola-mokhnach:
thanks a lot for your contribution @mgkeeley . I would be happy to review it, but my knowledge of C# is rather poor. Perhaps @GrzegorzSidwa could assist with a review?

PR #218, opened 2018-04-12
Pseudo-implemnenting IDisposable on AppiumCommandExecutor

jimevans:
## Change list

This commit implements the disposable pattern on AppiumCommandExecutor,
doing everything except actually declaring that the class implements
IDisposable. The upstream Selenium project will be changing the
ICommandExecutor interface to extend IDisposable in the near future, and
this change is required to prevent errors when that change is made.

## Types of changes

## Details

This PR implements the Disposable pattern as recommended by Microsoft for .NET code. The implementation includes a public `Dispose()` method and a protected `Dispose(bool)` method. As an added code cleanup, it also moves disposal of the service object into the Dispose method for the command executor class. If that change is deemed too risky, that portion of it may be easily reverted.

### DIFF
--- a/appium-dotnet-driver/Appium/Service/AppiumCommandExecutor.cs
+++ b/appium-dotnet-driver/Appium/Service/AppiumCommandExecutor.cs
@@ -23,6 +23,7 @@ namespace OpenQA.Selenium.Appium.Service
         private readonly AppiumLocalService Service;
         private readonly Uri URL;
         private readonly ICommandExecutor RealExecutor;
+        private bool isDisposed;
 
         private static ICommandExecutor CreateRealExecutor(Uri remoteAddress, TimeSpan commandTimeout)
         {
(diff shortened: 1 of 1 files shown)


jimevans:
I am about to make `ICommandExecutor` in the next version of Selenium. If this PR is not merged before that release, all Appium users who update to the next version of the Selenium package will be broken. Please consider merging and releasing a binary with this PR soon.

mykola-mokhnach:
@jimevans thanks a lot for your contribution. Please join our Appium channel on Slack. There people usually respond faster and it is also an appropriate place to discuss development issues. You could also be a project contributor with merge permissions if you have enough knowledge in C#/selenium API and have passion to do it.

jlipps:
I went ahead and merged this. I do not have the ability to release a new version of the client though. @TikhomirovSergey do you?

RoakyWood:
Is this in a release yet? If so, which version of Appium

akinsolb:
Not yet. It’ll ill be in v4.0.0.2-beta

PR #238, opened 2018-07-14
Small code quality improvements

UselessToucan:
## Change list

Replace `&` with `&&` and fix null check in `iOSLockDeviceTest.Cleanup`
 
## Types of changes

### DIFF
--- a/appium-dotnet-driver/Appium/PageObjects/ByFactory.cs
+++ b/appium-dotnet-driver/Appium/PageObjects/ByFactory.cs
@@ -108,7 +108,7 @@ namespace OpenQA.Selenium.Appium.PageObjects
                 return null;
 
             Attribute[] attributes = null;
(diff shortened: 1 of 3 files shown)


appium-dotnet-driver/Appium/PageObjects/ByFactory.cs
```diff
@@ -108,7 +108,7 @@ private static List<By> GetListOfComplexBys(bool useSequence, bool useAll, List<
                 return null;
 
             Attribute[] attributes = null;
-            if (upperPlatform.Equals(MobilePlatform.Android.ToUpper()) & (upperAutomation != null &&
+            if (upperPlatform.Equals(MobilePlatform.Android.ToUpper()) && (upperAutomation != null &&
```
SrinivasanTarget: There is a need to check both the conditions here that's why @TikhomirovSergey has put `&` over `&&`.

Issue #227, opened 2018-05-14
appium-dotnet-driver don't support .net core 2.0


ParisHuang:
Description
1. create .net core project.
2. install Appium.Webdriver package by Nuget.
3. there are warning information: "Package 'Appium.WebDriver 3.0.0.2' was restored using '.NETFramework, Version=v4.6.1' instead of the project target framework '.NETCoreApp, Version=v2.0', This package may not be fully compatible with your project."

Problem: appium-dotnet-driver don't support .net core 2.0 project.

nadvolod:
Having same issue. Any updates on this?

**The problem**
Appium doesn't work on .NET Core 2.0

**Environment**
Appium version (or git revision) that exhibits the issue: 3.0.0.2
Desktop OS/version used to run Appium: Windows 10 and Mac OS High Sierra
Real device or emulator/simulator: Real device
Details
I'm attempting to run a simple Appium test that works for me with .NET Framework, but doesn't work for me using .NET Core.

**Code To Reproduce Issue [ Good To Have ]**
Please supply your own Sauce user name and access key.

public void TestNativeAndroidApp(string testName)
        {
            var capabilities = new DesiredCapabilities();
            capabilities.SetCapability("platformName", "Android");
            capabilities.SetCapability("platformVersion", "7.1.1");
            //TODO first you must upload an app to Test Object so that you get your app key
            capabilities.SetCapability("testobject_api_key", "0D6C044F19D0442BA1E11C3FF087F6A6");
            capabilities.SetCapability("deviceName", "Google Pixel");
            capabilities.SetCapability("deviceOrientation", "portrait");
            capabilities.SetCapability("name", testName);
            var rdcUrl = "https://us1.appium.testobject.com/wd/hub";
            //TODO should be interchangeable to run on Windows/Mac
            var sauceUser = Environment.GetEnvironmentVariable("SAUCE_USERNAME", EnvironmentVariableTarget.User);
            var sauceKey = Environment.GetEnvironmentVariable("SAUCE_ACCESS_KEY", EnvironmentVariableTarget.User);
            var sauceUlr =
                $"https://{sauceUser}:{sauceKey}@ondemand.saucelabs.com:443/wd/hub";
            var driver = new AndroidDriver<IWebElement>(new Uri(rdcUrl), capabilities);
            driver.Quit();
        }

LGFox:
I managed to run it on MacOS but I compiled it for .net core manually.
[...]

Issue #191, opened 2017-10-06
Is it possible to convert appium-dotnet-driver to .Net Standard 2.0 Library?


owenniblock:
## Description
We'd like to use this library on a mac environment using .Net Core - however because it's built for the full .Net Framework - we can't run it using `dotnet` command line tools.

I've done a proof of concept locally and the only thing I had to change was the `Resource` file. I've basically hacked it for my local environment but I'll try and do a better version to send to you when I get time. However I've not run Integration tests against this - just proven that it builds and will launch an app.

## Environment

* Mac

owenniblock:
There's a first pass at a conversion here:
https://github.com/owenniblock/appium-dotnet-driver/tree/feature/dotnet-core-support

Obviously I don't know how you're building nuget etc but hopefully this will be useful. The integration_tests project is a .Net Core project and I can run the tests on my mac using `dotnet test` however I'm currently getting errors. I suspect some of this is my local set-up.

owenniblock:
Just a quick update on this. A large proportion of my errors were due to the removal of `PageFactory` on .Net Core as detailed [in the Selenium release notes](https://github.com/SeleniumHQ/selenium/blob/master/dotnet/CHANGELOG).

I've added some categories to make it easier to run and chase down where errors are genuine regression errors/set-up issues etc.

We could split the integration tests in to .Net Core and a further full .Net project to make sure the `PageFactory` functionality isn't broken.

owenniblock:
Just a quick poke on this. I don't want to spend any more time on my PoC if it's not useful to you guys but it seems like a good idea to me (and would be useful for a project I'm about to start working on).

Astro03:
Yeah, for some reason I think we are actually doing this conversion already. I don't see the branch though but I know both @TikhomirovSergey  and I talked about this. right? I'm not sure how far he is on the conversion.

owenniblock:
Perfect! I'd happily contribute if you need/want me to. I've got that PoC version working against a .Net Standard project locally and it seems pretty hopeful.

TikhomirovSergey:
@owenniblock The first beta version since Feb'17 has been released. Please be in touch. Any contributions are welcome.

jason-edstrom:
@TikhomirovSergey Where could one find the source for the .Net Standard beta version?  I would love to take a look and potentially contribute if I can.

mcopjan:
Any update on this please?

umbravi:
[...]

PR #258, opened 2018-09-13
Support Selenium v3.14

TroyWalshProf:
## Change list
Move to .Net Standard
Support Selenium 3.14 
 
## Types of changes

- [X ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Details
Make project .Net Standard.
- The will require those using the .Net Framework to update to at least 4.6.1 
Add AppiumOptions
- Mirror what Selenium's move away from desired capabilities
- See https://github.com/SeleniumHQ/selenium/issues/6278#issuecomment-420728084 for more info
Add AppiumCapabilities 
- Create Appium specific capability class


appium-dotnet-driver/Appium/AppiumDriver.cs
```diff
@@ -40,50 +40,50 @@ public abstract class AppiumDriver<W> : RemoteWebDriver, IExecuteMethod, IFindsB
...
+        public AppiumDriver(ICommandExecutor commandExecutor, ICapabilities AppiumOptions)
+            : base(commandExecutor, AppiumOptions)
         {
             AppiumCommand.Merge(commandExecutor.CommandInfoRepository);
             ElementFactory = CreateElementFactory();
         }
 
-        public AppiumDriver(ICapabilities desiredCapabilities)
-            : this(AppiumLocalService.BuildDefaultService(), desiredCapabilities)
+        public AppiumDriver(ICapabilities AppiumOptions)
```
mykola-mokhnach: should `appiumOptions` start with a capital letter everywhere?

TroyWalshProf:
I actually did think about that.  
From a maintenance perspective I think we need an AppiumOptions class. 
*I am confident that this code will need to be updated when DesiredCaps are no longer available.
That being said, if this change goes through I could see us extending it with iOS, Android, Windows and Tizen specific option classes. That way we get the best of both world.

akinsolb:
Sure. This takes a task out of my weekend. Descendants can always be done later if necessary., 
Did you manage to run a test against this branch? Atm, the CI doesn't have verification tests yet but running the last written test should sufficient, I think that would be the IOSLockDeviceTest in order to merge.

TroyWalshProf:
@jlipps - This is defiantly a breaking change.
1. Changes AppiumDriver (and iOS, Android, etc.) constructor signatures
2. Removes support for .Net 4.6 and below, but adds support for .Net Core, Mono, Xamarin and UWP

Moving over to use the DriverOptions is based off an [interaction ](https://github.com/SeleniumHQ/selenium/issues/6278#issuecomment-420728084) with Jim Evens.
*Seems like the best long term strategy for remaining compatible with future versions of Selenium.

jlipps:
@TroyWalshProf @akinsolb in that case, full steam ahead! Selenium and @jimevans compatibility are very important.

jimevans:
TL;DR I’m in favor of this PR, but some thought should be put into the design of this part of the Appium .NET API going forward. 

I’m an outsider who is offering an  unsolicited opinion on a project whose community I’m not an active member of. Take the below with a grain of salt.

Dropping support for .NET Framework and moving to .NET Core only at this time would be a mistake, in my mostly irrelevant opinion. As mentioned by @akinsolb, the versions of `PageFactory` that are currently available (formerly in `WebDriver.Support`, now deprecated and migrated to the DotNetSeleniumExtras project) are not available in .NET Standard/Core projects. 

Moreover, and this is simply a philosophical difference, this change appears to be simply substituting one “allow arbitrary names and values with no validity checking” property bag for another. The whole point of moving from `DesiredCapabilities` to a type-safe options class in the first place was to provide guidance and prevent user frustration because the capability being requested used the wrong (arbitrary string) name or the wrong (data type or invalid) value. Again, if the Appium project doesn’t care about that user experience, and it’s a perfectly legitimate choice not to, that’s a fine approach. However, I would put some thought into the API before making this the last word on the subject.

To forestall the inevitable question about why such a change is needed in the first place, correct use of DesiredCapabilities has been a source of confusion for users in all but the simplest of cases. It results in cargo cult, copy-paste code with no understanding of why the arbitrary capability is being set or if it’s even appropriate and supported by the target driver. Coupled with the W3C WebDriver Specification’s strictness around capabilities in the protocol, and the ability of implementations to reject connection requests with unknown or invalid capabilities, making the users’ and maintainers’ lives a little easier by preventing those types of errors in the first place seems like an obvious answer. 

Finally, and this is simply by way of communication with a large downstream project about future possibilities, I’m thinking about ways to make the browser-specific `DriverOptions` API clearer, and `AddAdditionalCapability` is likely to be changing, which probably means deprecation, to be replaced by a new method or methods. I need a better way to solve the problem where a user wants to add a capability which belongs to a browser-specific driver dictionary (e.g., “goog:chromeOptions”, “moz:firefoxOptions”, etc.) and a global or top-level capability (like SauceLabs or BrowserStack user and API key capabilities). Right now, specific options classes have an overload for the method, but this is neither discoverable nor elegant, and the size of the user base using the imprecise code makes it impossible to just change the behavior to something that feels more correct.

TroyWalshProf:
@akinsolb - Added .Net 4.5 support back.

TroyWalshProf:
@akinsolb - This change also addresses 
https://github.com/appium/appium-dotnet-driver/issues/227
https://github.com/appium/appium-dotnet-driver/issues/191
*.Net Core can use .Net Standard libraries

theAshenfire:
Hi,
Is there an ETA when this change will be available via NuGet?

akinsolb:
@theAshenfire This is [available](
https://www.nuget.org/packages/Appium.WebDriver/4.0.0.3-beta) in the beta version

jshinn6788:
What exactly is the resolution to the "Desired Capabilities" deprecated error?  because I'm still getting this error.

TroyWalshProf:
@jshinn6788 - Desired capabilities for Selenium has been deprecated.
What they did in the Selenium project was to  create driver specific options.
See: https://github.com/SeleniumHQ/selenium/blob/16559ddb469ec45bbdfef062a5638fb70cb18224/java/client/test/org/openqa/selenium/chrome/ChromeOptionsFunctionalTest.java#L89

This  pattern was followed in Appium.
Now instead of using Desired capabilities you use AppiumOptions.
See: https://github.com/appium/appium-dotnet-driver/blob/1dcb236576a92990c2580b0e600e6c9f3a841c23/test/integration/Windows/StickyNotesTest.cs#L53

jshinn6788:
Awesome!
This was very helpful
Thank you

Issue #251, opened 2018-08-10
Add support for screen recording endpoints


mykola-mokhnach:
https://github.com/appium/java-client/pull/814

jlipps:
@jamie-sherriff we'll need your e-mail somehow for the invite. you can DM me on twitter (@jlipps) if you don't want to put it here in public.

PR #269, opened 2018-10-13
Implement add support for screen recording endpoints #251

jamie-sherriff:
## Change list

Implemented add support for screen recording endpoints which covers #251

## Types of changes

## Details

Haven't added any iOS tests as i currently don't have a Mac OS dev environment setup. Is someone able to check that?

### DIFF
--- a/appium-dotnet-driver/Appium/AppiumDriver.cs
+++ b/appium-dotnet-driver/Appium/AppiumDriver.cs
@@ -432,14 +432,27 @@ namespace OpenQA.Selenium.Appium
         #endregion Session Data
 
         #region Device Methods
-        public void StartRecordingScreen()
+        /// <summary>
+        /// Rotates Device.
+        /// </summary>
+        /// <param name="opts">rotations options like the following:
+        /// new Dictionary<string, object> {{"x", 114}, {"y", 198}, {"duration", 5}, 
+        /// {"radius", 3}, {"rotation", 220}, {"touchCount", 2}}
+        /// </param>
+        public void StartRecordingScreen(Dictionary<string, object> opts)
         {
-            throw new NotImplementedException();
+            Dictionary<string, object> parameters = new Dictionary<string, object>();
+            foreach (KeyValuePair<string, object> opt in opts)
+            {
+                parameters.Add(opt.Key, opt.Value);
+            }
+            Execute(AppiumDriverCommand.StartRecordingScreen, parameters);
         }
 
-        public void StopRecordingScreen()
+        public String StopRecordingScreen()
         {
-            throw new NotImplementedException();
+            var commandResponse = ((IExecuteMethod)this).Execute(AppiumDriverCommand.StopRecordingScreen);
+            return commandResponse.ToString();
         }
 
         #endregion Device Methods
--- a/appium-dotnet-driver/Appium/AppiumDriver.cs
+++ b/appium-dotnet-driver/Appium/AppiumDriver.cs
@@ -452,7 +452,7 @@ namespace OpenQA.Selenium.Appium
         public String StopRecordingScreen()
         {
             var commandResponse = ((IExecuteMethod)this).Execute(AppiumDriverCommand.StopRecordingScreen);
-            return commandResponse.ToString();
+            return commandResponse.Value as string;
         }
 
         #endregion Device Methods
--- a/appium-dotnet-driver/Appium/AppiumDriver.cs
+++ b/appium-dotnet-driver/Appium/AppiumDriver.cs
@@ -37,7 +37,7 @@ namespace OpenQA.Selenium.Appium
     {
 
         private const string NativeApp = "NATIVE_APP";
-        
+
         #region Constructors
 
         public AppiumDriver(ICommandExecutor commandExecutor, ICapabilities appiumOptions)
@@ -430,6 +430,19 @@ namespace OpenQA.Selenium.Appium
                                  && Context.IndexOf(NativeApp, StringComparison.OrdinalIgnoreCase) < 0;
 
         #endregion Session Data
+
+        #region Device Methods
+        public void StartRecordingScreen()
+        {
+            throw new NotImplementedException();
+        }
+
+        public void StopRecordingScreen()
+        {
+            throw new NotImplementedException();
+        }
+
+        #endregion Device Methods
 
         #endregion Public Methods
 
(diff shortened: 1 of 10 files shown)


appium-dotnet-driver/Appium/AppiumDriver.cs
```diff
@@ -431,6 +429,38 @@ public object GetSessionDetail(string detail)
...
         #endregion Session Data
 
+        #region Device Methods
+
+        /// <summary>
+        /// Start recording the device screen
+        /// </summary>
+        /// <param name="opts">recording options like the following:
+        ///  Dictionary<string, object> StartRecordOptions = new Dictionary<string, object>();
+        ///  StartRecordOptions.Add("bit_rate", "1");
```
mykola-mokhnach: Please add a proper options documentation:

https://github.com/appium/appium-android-driver/blob/master/lib/commands/recordscreen.js
https://github.com/appium/appium-xcuitest-driver/blob/master/lib/commands/recordscreen.js

appium-dotnet-driver/Appium/AppiumDriver.cs
```diff
@@ -431,6 +429,38 @@ public object GetSessionDetail(string detail)
...
+        ///  StartRecordOptions.Add("video_size", "1280x720");
+        /// </param>
+        public void StartRecordingScreen(Dictionary<string, object> opts)
+        {
+            Dictionary<string, object> parameters = new Dictionary<string, object>();
```
mykola-mokhnach: why it is necessary to copy all values?

appium-dotnet-driver/Appium/AppiumDriver.cs
```diff
@@ -431,6 +429,38 @@ public object GetSessionDetail(string detail)
...
+                parameters.Add(opt.Key, opt.Value);
+            }
+            Execute(AppiumDriverCommand.StartRecordingScreen, parameters);
+        }
+
+
+        /// <summary>
+        /// Stop recording the device screen
+        /// </summary>
+        public String StopRecordingScreen()
```
mykola-mokhnach: stopRecording also supports options

appium-dotnet-driver/Appium/AppiumDriver.cs
jamie-sherriff: I copied the Rotate command here, 

```
 public void Rotate(Dictionary<string, int> opts)
        {
            Dictionary<string, object> parameters = new Dictionary<string, object>();
            foreach (KeyValuePair<string, int> opt in opts)
            {
                parameters.Add(opt.Key, opt.Value);
            }
            Execute(AppiumDriverCommand.Rotate, parameters);
        }

```


would you prefer to have a statically typed object? like the Location command?

appium-dotnet-driver/Appium/AppiumDriver.cs
jamie-sherriff: Sure thing, will add in.

appium-dotnet-driver/Appium/AppiumDriver.cs
jamie-sherriff: It takes URL parameters but not JSON parameter options which is what the command is referring to. i was going off http://appium.io/docs/en/commands/device/recording-screen/stop-recording-screen/

There do not appear to any existing commands passing through url parameters and it looks like its coming from the OpenQA.Selenium.Remote namespace? Are you able to help on how to pass them through?

appium-dotnet-driver/Appium/AppiumDriver.cs
jamie-sherriff: @mykola-mokhnach  just noticed the docs on http://appium.io/docs/en/commands/device/recording-screen/start-recording-screen/ don't match what you linked. The JSON parameters are underscore vs camelCase. Should i go by camelcase since that's what it was implemented in?

KazuCocoa: camelCase is correct. Let me update the doc. Thanks!
e.g. 
https://github.com/appium/ruby_lib_core/blob/master/test/unit/android/device/w3c/commands_test.rb#L229

jamie-sherriff:
@KazuCocoa  also just noticed while looking at more docs for bitRate is mentioned as iOS only but mentions the android API?

appium-dotnet-driver/Appium/AppiumDriver.cs
jamie-sherriff: Have now assumed stopRecording takes JSON parameters not URL query paramaters which is what it looks like other client libraries are doing.

appium-dotnet-driver/Appium/RecordScreen.cs
```diff
@@ -0,0 +1,158 @@
...
+        /// Whether to ignore the result of previous capture and start a new recording
+        /// immediately (`True` value). By default  (`False`) the endpoint will try to catch and return the result of
+        /// the previous capture if it's still available.
+        /// </summary>
+        [JsonProperty(PropertyName = "forceRestart")]
+        public bool? ForceRestart;
+
+        /// <summary>
+        /// The actual time limit of the recorded video in seconds.
+        /// The default value for both iOS and Android is 180 seconds (3 minutes).
```
mykola-mokhnach: This is changed since Appium 1.9.1 and it is possible to set higher values, but it requires ffmpeg binary to be present in PATH

appium-dotnet-driver/Appium/RecordScreen.cs
```diff
@@ -0,0 +1,158 @@
...
+        [JsonProperty(PropertyName = "videoSize")]
+        public string VideoSize;
+
+        /// <summary>
+        /// Makes the recorder to display an additional information on the video overlay,
+        /// such as a timestamp, that is helpful in videos captured to illustrate bugs.
+        /// This option is only supported since API level 27 (Android P).
+        /// </summary>
+        [JsonProperty(PropertyName = "bugReport")]
+        public string BugReport;
```
mykola-mokhnach: this is a boolean

appium-dotnet-driver/Appium/RecordScreen.cs
```diff
@@ -0,0 +1,158 @@
...
+
+        /// <summary>
+        /// The video bit rate for the video, in megabits per second.
+        /// The default value is 4. You can increase the bit rate to improve video quality,
+        /// but doing so results in larger movie files.
+        /// </summary>
+        [JsonProperty(PropertyName = "bitRate")]
+        public string BitRate;
```
mykola-mokhnach: this is an integer

Issue #237, opened 2018-07-08
Add support for FindsByWindowsAutomationAttribute


cosminstirbu:
## Description

We have a suite of tests that need to run on both iOS and UWP.

We're using the `AppiumPageObjectMemberDecorator` to decorate our elements, but unfortunately there's no attribute for Windows.

Currently the `AppiumPageObjectMemberDecorator` and `ByFactory` classes only support iOS and Android.

It would be useful to support Windows as well - maybe by introducing a `FindsByWindowsAutomationAttribute`.

Until this is done - is there maybe a workaround that we could use to decorate our elements for both iOS and UWP?

Thanks,
Cosmin

cosminstirbu:
I've forked the repo and made the changes as a patch against 3.0.0.2 (since this is the version we're currently using)

You can find the changes here: https://github.com/cosminstirbu/appium-dotnet-driver/commit/5fbd7279b281efaddca0a3932037cdb40e404c01

If you're happy with them, I can make the same changes against `master` as well and open a PR.

Thanks,
Cosmin

cosminstirbu:
After adding this feature, I've ran into this issue https://github.com/Microsoft/WinAppDriver/issues/454

Any help would be appreciated.

akinsolb:
@cosminstirbu make a pull request

cosminstirbu:
I'll do a rebase sometime by the end of this week and I'll create the pull request this weekend hopefully.

akinsolb:
Closed by #276

PR #276, opened 2018-10-23
Add support for FindsByWindowsAutomationAttribute

cosminstirbu:
Add support for FindsByWindowsAutomationAttribute - see https://github.com/appium/appium-dotnet-driver/issues/237 for more details.

P.S - The name of the branch is a bit misleading since I've rebased it against master (4.0.0.0 beta 4).

### DIFF
--- a/src/Appium.Net/Appium/Windows/WindowsDriver.cs
+++ b/src/Appium.Net/Appium/Windows/WindowsDriver.cs
@@ -13,6 +13,7 @@
 //limitations under the License.
 
 using System;
+using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using OpenQA.Selenium.Appium.Enums;
 using OpenQA.Selenium.Appium.Interfaces;
@@ -109,6 +110,25 @@ namespace OpenQA.Selenium.Appium.Windows
             : base(service, SetPlatformToCapabilities(AppiumOptions, Platform), commandTimeout)
         {
         }
+
+        #region Contex
+
+        public override string Context
+        {
+            get => "NATIVE_APP";
+            set => throw new NotImplementedException();
+        }
+
+        public override ReadOnlyCollection<string> Contexts
+        {
+            get
+            {
+                var contexts = new List<string> { Context };
+                return contexts.AsReadOnly();
+            }
+        }
+
+        #endregion Context
 
         #region IFindByWindowsUIAutomation Members
 
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -309,7 +309,7 @@ namespace OpenQA.Selenium.Appium
 
         #region Context
 
-        public string Context
+        public virtual string Context
         {
             get
             {
@@ -323,7 +323,7 @@ namespace OpenQA.Selenium.Appium
             }
         }
 
-        public ReadOnlyCollection<string> Contexts
+        public virtual ReadOnlyCollection<string> Contexts
         {
             get
             {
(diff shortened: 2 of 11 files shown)


akinsolb:
@cosminstirbu Are you able to provide some integration tests? in order for it to be merged

cosminstirbu:
Unfortunately I'm quite caught up so I won't be able to do it in the near future (2-4 weeks).

If you're OK with it, please merge it as is, and I'll revisit it at some point, either me or someone from my team, as I think it would be a good "exercise" for them to contribute on an open source project.

cosminstirbu:
@akinsolb 

I've updated the existing Windows tests because they were failing due to outdated accessibility identifiers.

I've also added an integration tests for the Windows Page Object attributes that I've added.

Please review and let me know if there's anything else that you'd like me to add.

akinsolb:
@cosminstirbu I'll try get this done today.  I had build server issues whilst testing last year.

src/Appium.Net/Appium/Windows/WindowsDriver.cs
```diff
@@ -110,6 +111,31 @@ public WindowsDriver(Uri remoteAddress, AppiumOptions AppiumOptions, TimeSpan co
         {
         }
 
+        #region Contex
+
+        public override string Context
+        {
+            get
+            {
+                return "NATIVE_APP";
```
KazuCocoa: - Does windows driver have no implementation for `get context` as well?
- Does windows driver raise `InvalidOperationException` for `set`? If so, we simply though the error to users.
    - We can reduce client-side implementation

cosminstirbu: WinAppDriver returns Unknown Command as it can be seen here https://gist.github.com/cosminstirbu/14375a1f101ad5fed16648a37b746102

On the client side client, the exception is thrown in `ContentMappedBy` via `GetReturnRelevantBys`. I could do a try / catch there on the catch clause `return map[ContentTypes.NATIVE];`

cosminstirbu: However this would "mess up" the logs since this call will be frequently logged :-?

KazuCocoa: > However this would "mess up" the logs since this call will be frequently logged :-?

It's trade-off thing, I would say.
We are implementing clients to throw errors to users such as `not implemented yet` without modification in clients layer as possible. We can fix the error only updating Appium server side. We do not need to force users to update frequently.

I am not familiar with windriver's situation. Thus, I'm fine to go this way 👍if it is better for windows driver.

src/Appium.Net/Appium/Windows/WindowsDriver.cs
```diff
@@ -110,6 +111,31 @@ public WindowsDriver(Uri remoteAddress, AppiumOptions AppiumOptions, TimeSpan co
         {
         }
 
+        #region Contex
+
+        public override string Context
+        {
+            get
```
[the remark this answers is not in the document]
cosminstirbu: Done

src/Appium.Net/Appium/Windows/WindowsDriver.cs
```diff
@@ -110,6 +111,31 @@ public WindowsDriver(Uri remoteAddress, AppiumOptions AppiumOptions, TimeSpan co
...
+            {
+                return "NATIVE_APP";
+            }
+            set
```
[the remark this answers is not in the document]
cosminstirbu: Done

cosminstirbu:
@akinsolb I have ran the tests without setting the remoteFlag, I have used only the local service without any other changes (since I have Appium server installed locally, and it comes with WinAppDriver out of the box, so no other changes needed there)

I've copy pasted the setup from the existing tests, so I'm not sure if the existing remote infrastructure is working or not.

akinsolb:
Ok. The reason I asked is because the context endpoint isn't supported by the WinApp Driver. I was using the standalone while testing the context, but appium server side handles the getContext endpoint.

All LGTM

PR #278, opened 2018-10-24
Support clipboard endpoint

akinsolb:
## Change list

- IOSDriver
- AndroidDriver

## Types of changes

## Details
Enable users to utilise the SetClipboard and GetClipboard features on Android and IOS platform.

### DIFF
--- a/appium-dotnet-driver/Appium/Android/AndroidDriver.cs
+++ b/appium-dotnet-driver/Appium/Android/AndroidDriver.cs
@@ -257,12 +257,25 @@ namespace OpenQA.Selenium.Appium.Android
         }
 
         /// <inheritdoc />
-        public void SetClipboard(ClipboardContentType contentType, string base64Content) =>
-            AppiumCommandExecutionHelper.SetClipboard(this, contentType, base64Content);
+        public void SetClipboard(ClipboardContentType contentType, string base64Content)
+        {
+            switch (contentType)
+            {
+                case ClipboardContentType.Image:
+                case ClipboardContentType.Url:
+                    throw new NotImplementedException($"Android only supports contentType: {nameof(ClipboardContentType.PlainText)}");
+                case ClipboardContentType.PlainText:
+                    AppiumCommandExecutionHelper.SetClipboard(this, contentType, base64Content);
+                    break;
+                default:
+                    AppiumCommandExecutionHelper.SetClipboard(this, contentType, base64Content);
+                    break;
+            }
+        }
 
         /// <inheritdoc />
         public void SetClipboardText(string textContent, string label) =>
-            AppiumCommandExecutionHelper.SetClipboard(this, ClipboardContentType.PlainText, label);
+            AppiumCommandExecutionHelper.SetClipboardText(this, ClipboardContentType.PlainText, textContent, label);
 
         /// <inheritdoc />
         public string GetClipboard(ClipboardContentType contentType)
@@ -271,11 +284,11 @@ namespace OpenQA.Selenium.Appium.Android
             {
                 case ClipboardContentType.Image:
                 case ClipboardContentType.Url:
-                    throw new NotImplementedException($"Android only supports {nameof(ClipboardContentType.PlainText)}");
+                    throw new NotImplementedException($"Android only supports contentType: {nameof(ClipboardContentType.PlainText)}");
                 case ClipboardContentType.PlainText:
-                    return GetClipboardText();
+                    return AppiumCommandExecutionHelper.GetClipboard(this, contentType);
                 default:
-                    return GetClipboardText();
+                    return AppiumCommandExecutionHelper.GetClipboard(this, contentType);
             }
         }
 
--- a/appium-dotnet-driver/Appium/Android/AndroidDriver.cs
+++ b/appium-dotnet-driver/Appium/Android/AndroidDriver.cs
@@ -257,12 +257,12 @@ namespace OpenQA.Selenium.Appium.Android
         }
 
         /// <inheritdoc />
-        public void SetClipboard(byte[] base64Content, ClipboardContentType contentType, string label = null) =>
-            AppiumCommandExecutionHelper.SetClipboard(this, contentType, base64Content, label);
+        public void SetClipboard(ClipboardContentType contentType, string base64Content) =>
+            AppiumCommandExecutionHelper.SetClipboard(this, contentType, base64Content);
 
         /// <inheritdoc />
         public void SetClipboardText(string textContent, string label) =>
-            AppiumCommandExecutionHelper.SetClipboardText(this, ClipboardContentType.PlainText, textContent, label);
+            AppiumCommandExecutionHelper.SetClipboard(this, ClipboardContentType.PlainText, label);
 
         /// <inheritdoc />
         public string GetClipboard(ClipboardContentType contentType)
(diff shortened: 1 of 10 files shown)


appium-dotnet-driver/Appium/Android/AndroidDriver.cs
```diff
@@ -255,5 +255,44 @@ public ConnectionType ConnectionType
...
+        /// <inheritdoc />
+        public void SetClipboard(ClipboardContentType contentType, string base64Content)
+        {
+            switch (contentType)
+            {
+                case ClipboardContentType.Image:
+                case ClipboardContentType.Url:
+                    throw new NotImplementedException($"Android only supports contentType: {nameof(ClipboardContentType.PlainText)}");
+                case ClipboardContentType.PlainText:
+                    AppiumCommandExecutionHelper.SetClipboard(this, contentType, base64Content);
```
mykola-mokhnach: should it call SetClipboardText?

appium-dotnet-driver/Appium/Android/AndroidDriver.cs
```diff
@@ -255,5 +255,44 @@ public ConnectionType ConnectionType
...
+        /// <inheritdoc />
+        public string GetClipboard(ClipboardContentType contentType)
+        {
+            switch (contentType)
+            {
+                case ClipboardContentType.Image:
+                case ClipboardContentType.Url:
+                    throw new NotImplementedException($"Android only supports contentType: {nameof(ClipboardContentType.PlainText)}");
+                case ClipboardContentType.PlainText:
+                    return AppiumCommandExecutionHelper.GetClipboard(this, contentType);
```
mykola-mokhnach: should it call GetClipboardText

appium-dotnet-driver/Appium/AppiumCommand.cs
```diff
@@ -32,8 +32,15 @@ public class AppiumCommand
 
             #endregion Context Commands
 
-            #region Appium Commands
+            #region  Driver Commands
```
mykola-mokhnach: extra space here?

appium-dotnet-driver/Appium/AppiumCommandExecutionHelper.cs
```diff
@@ -58,6 +65,77 @@ public static void HideKeyboard(IExecuteMethod executeMethod, string strategy =
...
+            switch (clipboardContentType)
+            {
+                case ClipboardContentType.Url:
+                case ClipboardContentType.PlainText:
+                    var encodedContentBytes = Encoding.UTF8.GetBytes(base64Content);
+                    base64Content = Convert.ToBase64String(encodedContentBytes);
+                    break;
+            }
+            return (string)executeMethod.Execute(AppiumDriverCommand.SetClipboard,
+                PrepareArguments(new[] { "content", "contentType", "label" },
```
mykola-mokhnach: `label` can simply be skipped. There is no need to set it to null

appium-dotnet-driver/Appium/AppiumCommandExecutionHelper.cs
```diff
@@ -58,6 +65,77 @@ public static void HideKeyboard(IExecuteMethod executeMethod, string strategy =
...
+                PrepareArgument("contentType", clipboardContentType.ToString().ToLower())).Value;
+        }
+
+        public static string GetClipboardText(IExecuteMethod executeMethod)
+        {
+            var encodedContentBytes = Convert.FromBase64String(GetClipboard(executeMethod, ClipboardContentType.PlainText));
+            return Encoding.UTF8.GetString(encodedContentBytes);
+        }
+
+        public static string SetClipboard(IExecuteMethod executeMethod, ClipboardContentType clipboardContentType, string base64Content)
```
mykola-mokhnach: I find the design of this method is not very obvious since `base64Content` can be an actual base64-encoded content or a normal string

mykola-mokhnach: I'd also expect urls to be represented by special type instead of raw strings

appium-dotnet-driver/Appium/AppiumCommandExecutionHelper.cs
```diff
@@ -58,6 +65,77 @@ public static void HideKeyboard(IExecuteMethod executeMethod, string strategy =
...
+
+        /// <summary>
+        /// Prepares single command argument
+        /// </summary>
+        /// <param name="name">The parameter name</param>
+        /// <param name="value">The parameter value</param>
+        /// <returns></returns>
+        internal static Dictionary<string, object> PrepareArgument(string name, object value)
+        {
+            var parameterBuilder = new Dictionary<string, object> { { name, value } };
```
mykola-mokhnach: do we need the intermediate `parameterBuilder` variable?

appium-dotnet-driver/Appium/AppiumCommandExecutionHelper.cs
```diff
@@ -58,6 +65,77 @@ public static void HideKeyboard(IExecuteMethod executeMethod, string strategy =
...
+        /// <summary>
+        /// Prepares a collection of command arguments
+        /// </summary>
+        /// <param name="names">The array of parameter names</param>
+        /// <param name="values">The array of parameter values</param>
+        /// <returns></returns>
+        internal static Dictionary<string, object> PrepareArguments(string[] names, object[] values)
+        {
+            var parameterBuilder = new Dictionary<string, object>();
+            for (var i = 0; i < names.Length; i++)
```
mykola-mokhnach: it probably makes sense to assert len(names) == len(values) first

appium-dotnet-driver/Appium/AppiumCommandExecutionHelper.cs
```diff
@@ -58,6 +65,77 @@ public static void HideKeyboard(IExecuteMethod executeMethod, string strategy =
...
+            {
+                if (names[i] != string.Empty && values[i] != null)
```
mykola-mokhnach: shouldn't null values be transformed to JSON null instead?

appium-dotnet-driver/Appium/Enums/ClipboardContentType.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace OpenQA.Selenium.Appium.Enums
+{
+    public enum ClipboardContentType
+    {
+        PlainText,
+        Image,
```
mykola-mokhnach: Is setting of images supported?

akinsolb: Only for IOS. See the commits made in the integration_test project

appium-dotnet-driver/Appium/Android/AndroidDriver.cs
akinsolb: I had it call setClipboard in an earlier commit. I think a test was failing, which I then fixed. Let me check

appium-dotnet-driver/Appium/AppiumCommandExecutionHelper.cs
akinsolb: Are you saying the label field is optional? if so, If it is null, it shouldn't be sent to the server

appium-dotnet-driver/Appium/Android/AndroidDriver.cs
akinsolb: Not sure, but I think other clients allow users to receive a base64 string for PlainText. I left the GetClipboard and SetClipboard methods to do this, with the Set/GetClipboardText methods to return the decoded string. What do you think?

appium-dotnet-driver/Appium/Android/AndroidDriver.cs
mykola-mokhnach: this clause might be merged with default to reduce duplication. Same for SetClipboard

appium-dotnet-driver/Appium/AppiumCommandExecutionHelper.cs
akinsolb: I believe this needs to be a string in order to support .NetStandard, as there is no URL type.

appium-dotnet-driver/Appium/AppiumCommandExecutionHelper.cs
akinsolb: No. That serialisation is handled by Selenium, and works very well

PR #292, opened 2018-11-19
BackgroundApp update

AnnaWyrwal:
## Change list
Update of BackgroundApp method
New helper method PrepareArgument introduced
## Types of changes

## Integration tests
- [ ] Have you provided integration tests to pass against the latest version of appium? (for Bugfix or New feature)

## Details
Update of BackgroundApp method due to change in Appium, which introduced new way of handling parameters passed to BackgroundApp method and marked old approach as deprecated (see [documentation](https://appium.readthedocs.io/en/latest/en/commands/device/app/background-app/ "Background app")). 
Current Appium requires object containing {["timeout"]: seconds} to be passed as value of "seconds" argument. Passed value can be null/negative number to deactivate app or positive integer designating how long, in seconds, to background app for. 
To achieve it with more readability new method PrepareArgument (parallel to method used in Java client) was introduced for preparing JSON objects. Existing methods was adjusted to use it as well. 

Updated BackgroundApp method allows to:
- put app into the background for n seconds by:
`driver.BackgroundApp(n) `
- deactivate app (aka Home button) by:
`driver.BackgroundApp(-1)`
or 
`driver.BackgroundApp(null)`

Newly introduced helper method PrepareArguments allows to:
- prepare simple JSON objects:
  `PrepareArgument("appPath", pathToApp) ` 
is equivalent for: 
`  "appPath": pathToApp`
 - build complex JSON objects by recursive calls:
`PrepareArgument("seconds", PrepareArgument("timeout", seconds))`
is equivalent for : 
```
  "seconds": {
     "timeout": seconds
   }
```

### DIFF
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -23,6 +23,7 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
+using System.Runtime.ConstrainedExecution;
 using Castle.Core.Internal;
 using OpenQA.Selenium.Appium.ScreenRecording;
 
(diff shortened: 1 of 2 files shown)


src/Appium.Net/Appium/AppiumDriver.cs
```diff
@@ -473,6 +468,11 @@ internal static ICapabilities SetPlatformToCapabilities(DriverOptions dc, string
             return result.AsReadOnly();
         }
 
+        protected static Dictionary<string, object> PrepareArgument(string param, object value)
```
mykola-mokhnach: is it possible to have a method with variable number of params and values?

src/Appium.Net/Appium/Interfaces/IInteractsWithApps.cs
```diff
@@ -40,10 +40,10 @@ public interface IInteractsWithApps : IExecuteMethod
         void ResetApp();
 
         /// <summary>
-        /// Backgrounds the current app for the given number of seconds.
+        /// Backgrounds the current app for the given number of seconds or deactivates app completely if negative number or null is given. 
         /// </summary>
-        /// <param name="seconds">a string containing the number of seconds.</param>
-        void BackgroundApp(int seconds);
+        /// <param name="seconds">a integer number containing the number of seconds or null.</param>
+        void BackgroundApp(int? seconds);
```
mykola-mokhnach: passing null may be confusing. I'd rather add an overridden version of method without arguments that minimises the app

AnnaWyrwal: You're right. I don't know why I didn't thought about it! I'll change that :)  👍

src/Appium.Net/Appium/AppiumDriver.cs
AnnaWyrwal: Do you mean sth like:
`PrepareArguments(key1, value1, key2, value2, ...)`?

mykola-mokhnach: yes.  An alternative to `ImmutableMap.of(....)` in Google Guava's implementation

AnnaWyrwal: Then yes, I think it's possible. I'll try to do it

src/Appium.Net/Appium/Interfaces/IInteractsWithApps.cs
AnnaWyrwal: Done

src/Appium.Net/Appium/AppiumDriver.cs
AnnaWyrwal: @akinsolb You're right! Awesome! Then I can use it and remove my method. Nice. On it.

AnnaWyrwal: Done

PR #303, opened 2018-12-27
Adds functionality that allows redirecting the AppiumLocalService process output to Console to allow easier debugging

cosminstirbu:
## Change list

Added capability to redirect the Appium Local Service output similar to the way Java client does this - https://github.com/appium/java-client/blob/98daabe3943b606c5416818b73add493d889cc94/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java#L212
 
## Types of changes

## Integration tests
- [x] Have you provided integration tests to pass against the latest version of appium? (for Bugfix or New feature)

## Details

Even though the name of the branch suggests that we're redirecting the Error as well, it doesn't really make sense to do so (this was my initial intention - hence the name of the branch).

As a difference between the .NET client and the Java client, I have chosen not to redirect to `Console.Out` by default (see https://github.com/appium/java-client/blob/98daabe3943b606c5416818b73add493d889cc94/src/main/java/io/appium/java_client/service/local/AppiumDriverLocalService.java#L64) since there are issues in the way some xUnit frameworks (such as `NUnit` handle `Console.Out` see https://github.com/nunit/nunit/issues/2182#issuecomment-445453915)

Also, the output contains some "strange" characters, probably used by some other smarter consoles to color some ranges of the text:

`\u001b[35m[Appium]\u001b[39m Welcome to Appium v1.9.1`

Any recommendation on how we could sanitize this output, would be really appreciated.

Here you can also find a screenshot of out the output looks in VS 2017 - https://imgur.com/a/O8SMk52

### DIFF
--- a/src/Appium.Net/Appium/Service/AppiumLocalService.cs
+++ b/src/Appium.Net/Appium/Service/AppiumLocalService.cs
@@ -37,7 +37,11 @@ namespace OpenQA.Selenium.Appium.Service
         /// <returns>An instance of AppiumLocalService without special settings</returns>
         public static AppiumLocalService BuildDefaultService() => new AppiumServiceBuilder().Build();
 
-        internal AppiumLocalService(FileInfo nodeJS, string arguments, IPAddress ip, int port,
+        internal AppiumLocalService(
+            FileInfo nodeJS, 
+            string arguments, 
+            IPAddress ip, 
+            int port,
             TimeSpan initializationTimeout)
         {
             NodeJS = nodeJS;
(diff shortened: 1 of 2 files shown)


akinsolb:
@cosminstirbu User would probably have to add the --log-no-colors flag to the server arguments

cosminstirbu:
In case any one else is looking into disabling the colors, the correct way is to pass the --log-no-colors flag without any other value.

So `new OpenQA.Selenium.Appium.Service.Options.OptionCollector().AddArguments(new System.Collections.Generic.KeyValuePair<string, string>("--log-no-colors", string.Empty))` instead of 
`new Options.OptionCollector().AddArguments(new System.Collections.Generic.KeyValuePair<string, string>("--log-no-colors", "true"))`

PR #324, opened 2019-04-12
Add fingerprint endpoint

dpgraham:
## Change list

Add the [fingerprint](http://appium.io/docs/en/commands/device/authentication/finger-print/) command to the .NET Driver
 
## Types of changes

## Integration tests
- [x] Have you provided integration tests to pass against the latest version of appium? (for Bugfix or New feature)


test/integration/Android/FingerprintTest.cs
```diff
@@ -0,0 +1,48 @@
...
+    {
+        private AndroidDriver<IWebElement> _driver;
+        private const string ClipboardTestString = "Hello Clipboard";
+        private const string Base64RegexPattern = @"^[a-zA-Z0-9\+/]*={0,2}$";
+
+        [OneTimeSetUp]
+        public void BeforeAll()
+        {
+            var capabilities = Env.ServerIsRemote()
+                ? Caps.GetAndroidCaps(Apps.Get("androidApiDemos"))
```
mykola-mokhnach: the condition here has no effect, since both branches are equal

test/integration/Android/FingerprintTest.cs
```diff
@@ -0,0 +1,48 @@
...
+        [TearDown]
+        public void TearDown()
+        {
+            _driver?.CloseApp();
+        }
+
+        [Test]
+        public void TestSendFingerprint()
+        {
+            _driver.FingerPrint(1);
```
mykola-mokhnach: should we assert something here?

src/Appium.Net/Appium/AppiumDriver.cs
```diff
@@ -305,7 +308,7 @@ public Location Location
                 Execute(AppiumDriverCommand.SetLocation, location.ToDictionary());
             }
         }
-
+
```
mykola-mokhnach: this is not needed

test/integration/Android/FingerprintTest.cs
dpgraham: I don't know what to assert.... there's nothing in the apidemos app that detects fingerprints.

PR #340, opened 2019-08-13
Add Terminate App

broetchenrackete36:
## Change list

Add support for TerminateApp
 
## Types of changes

## Integration tests
- [ ] Have you provided integration tests to pass against the latest version of appium? (for Bugfix or New feature)

## Details

Added Terminate-App-Command (http://appium.io/docs/en/commands/device/app/terminate-app/).

### DIFF
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -227,10 +227,14 @@ namespace OpenQA.Selenium.Appium
         public void RemoveApp(string appId) =>
             Execute(AppiumDriverCommand.RemoveApp, AppiumCommandExecutionHelper.PrepareArgument("appId", appId));
 
-        public void TerminateApp(string appId, int? timeout = null) =>
-            Execute(AppiumDriverCommand.TerminateApp, 
+        public bool TerminateApp(string appId) =>
+            Convert.ToBoolean(Execute(AppiumDriverCommand.TerminateApp,
+                AppiumCommandExecutionHelper.PrepareArgument("appId", appId)).Value.ToString());
+
+        public bool TerminateApp(string appId, int timeout) =>
+            Convert.ToBoolean(Execute(AppiumDriverCommand.TerminateApp, 
                 AppiumCommandExecutionHelper.PrepareArguments(new string[] { "appId", "options" }, 
-                    new object[] { appId, new Dictionary<string, int?>() { { "timeout", timeout } } }));
+                    new object[] { appId, new Dictionary<string, object>() { { "timeout", timeout } } })).Value.ToString());
 
         public bool IsAppInstalled(string bundleId) =>
             Convert.ToBoolean(Execute(AppiumDriverCommand.IsAppInstalled,
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -227,8 +227,9 @@ namespace OpenQA.Selenium.Appium
         public void RemoveApp(string appId) =>
             Execute(AppiumDriverCommand.RemoveApp, AppiumCommandExecutionHelper.PrepareArgument("appId", appId));
 
-        public void TerminateApp(string appId) =>
-            Execute(AppiumDriverCommand.TerminateApp, AppiumCommandExecutionHelper.PrepareArgument("appId", appId));
+        public void TerminateApp(string appId, int? timeout = null) =>
+            Execute(AppiumDriverCommand.TerminateApp, 
+                AppiumCommandExecutionHelper.PrepareArguments(new string[] { "appId", "options" }, new object[] { appId, timeout }));
 
         public bool IsAppInstalled(string bundleId) =>
             Convert.ToBoolean(Execute(AppiumDriverCommand.IsAppInstalled,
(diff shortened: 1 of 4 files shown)


src/Appium.Net/Appium/AppiumDriver.cs
```diff
@@ -227,6 +227,10 @@ public void Rotate(Dictionary<string, int> opts)
         public void RemoveApp(string appId) =>
             Execute(AppiumDriverCommand.RemoveApp, AppiumCommandExecutionHelper.PrepareArgument("appId", appId));
 
+        public void TerminateApp(string appId, int? timeout = null) =>
+            Execute(AppiumDriverCommand.TerminateApp, 
+                AppiumCommandExecutionHelper.PrepareArguments(new string[] { "appId", "options" }, new object[] { appId, timeout }));
```
mykola-mokhnach: `timeout` should be a dictionary key

broetchenrackete36: I'm probably missing something obvious here, but I'm not exactly sure what you mean... Could you elaborate?

mykola-mokhnach: `options` argument is expected to be a dictionary, where `timeout` is just one of the items. Although, `timeout` is being passed as `flat` value, which is not going to work

broetchenrackete36: Ah, I see. Changed it to a dictionary.

src/Appium.Net/Appium/AppiumDriver.cs
```diff
@@ -227,6 +227,11 @@ public void Rotate(Dictionary<string, int> opts)
         public void RemoveApp(string appId) =>
             Execute(AppiumDriverCommand.RemoveApp, AppiumCommandExecutionHelper.PrepareArgument("appId", appId));
 
+        public void TerminateApp(string appId, int? timeout = null) =>
+            Execute(AppiumDriverCommand.TerminateApp, 
+                AppiumCommandExecutionHelper.PrepareArguments(new string[] { "appId", "options" }, 
+                    new object[] { appId, new Dictionary<string, int?>() { { "timeout", timeout } } }));
```
mykola-mokhnach: the timeout should not be included if it is not set

mykola-mokhnach: Generally the dictionary item types should be `<String, Any?>`

src/Appium.Net/Appium/Interfaces/IInteractsWithApps.cs
```diff
@@ -56,6 +56,13 @@ public interface IInteractsWithApps : IExecuteMethod
         /// <param name="appPath">a string containing the id of the app.</param>
         void RemoveApp(string appId);
 
+        /// <summary>
+        /// Terminates an App.
+        /// </summary>
+        /// <param name="appId">a string containing the id of the app.</param>
+        /// <param name="timeout">an integer number containing the retry-timeout</param>
```
mykola-mokhnach: this timeout means how long the application should take to be fully terminated. If it needs more time then the function will return false (500ms by default)

mykola-mokhnach: https://github.com/appium/java-client/blob/6d408e8011e3e81f6d23015876d81760b496f60e/src/main/java/io/appium/java_client/android/appmanagement/AndroidTerminateApplicationOptions.java

src/Appium.Net/Appium/AppiumDriver.cs
broetchenrackete36: Would you suggest an overload like for BackgroundApp instead of `int? timeout = null`?

mykola-mokhnach: Unfortunately I cannot advice nothing on this, since I don't mental C# client. Prefer the way, which is already common for other methods, so they all follow the single style.

src/Appium.Net/Appium/Interfaces/IInteractsWithApps.cs
```diff
@@ -56,6 +56,21 @@ public interface IInteractsWithApps : IExecuteMethod
...
+        /// <return>a boolean indicating if the app was terminated.</return>
+        bool TerminateApp(string appId);
+
+        /// <summary>
+        /// Terminates an App.
+        /// </summary>
+        /// <param name="appId">a string containing the id of the app.</param>
+        /// <param name="timeout">an integer number containing the timeout in ms for how long to wait until the application is terminated.</param>
+        /// <return>a boolean indicating if the app was terminated in the given timeout.</return>
+        bool TerminateApp(string appId, int timeout);
```
mykola-mokhnach: Does C# have a special type for Duration, like Java does? I would prefer to use it over the generic int

broetchenrackete36: Yes, TimeSpan. I haven't used it because the timeout for BackgroundApp is an int, although here it is in seconds and -1 deactivates the app. The Timeout for the TouchAction "wait" is in ms as long (in the java client both use a duration).

So I'm not sure what to use here, I think I'm leaning towards TimeSpan...

src/Appium.Net/Appium/Interfaces/IInteractsWithApps.cs
```diff
@@ -56,6 +58,21 @@ public interface IInteractsWithApps : IExecuteMethod
...
+        /// </summary>
+        /// <param name="appId">a string containing the id of the app.</param>
+        /// <return>a boolean indicating if the app was terminated.</return>
+        bool TerminateApp(string appId);
+
+        /// <summary>
+        /// Terminates an App.
+        /// </summary>
+        /// <param name="appId">a string containing the id of the app.</param>
+        /// <param name="timeout">an integer number containing the timeout in ms for how long to wait until the application is terminated.</param>
```
mykola-mokhnach: please update the docstring

mykola-mokhnach:
I like the nickname :)

Issue #309, opened 2019-01-30
appium 4.0.0.5-beta BackgroundApp(TIME) doesn't wait in background expected TIME return live immediately


vatsikmax:
## Description

appium 4.0.0.5-beta BackgroundApp(TIME) function doesn't wait in background expected TIME return live immediately

## Environment

* Appium nuget package version 4.0.0.5-beta:
* Appium server version 1.9.1:
* Node.js version 8.11.2:
* Android/8.1.0:
* Real device:

## Code To Reproduce Issue [ Good To Have ]

appiumDriver.BackgroundApp(10)

## Exception stacktraces

NO exception

## Link to Appium logs

https://gist.github.com/vatsikmax/e83998c614f86ccebc0be8ee09a8d9f4

KazuCocoa:
In android case, the payload should be `{"seconds": 10}`. Not `{"seconds":{"timeout":10}}`
Maybe, this is the cause...
(The latter case is for xcuitest)

matthew-horrocks:
Just been looking at this.

What's the best way to distinguish between Android` xcuitest` and `uiautomator2` in this method?

https://github.com/appium/appium-dotnet-driver/blob/4ec4c8b5447b7b5052f2b9c3a1a848ad62def410/src/Appium.Net/Appium/AppiumDriver.cs#L268

Presumably `uiautomator2` should be;

`Execute(AppiumDriverCommand.BackgroundApp,
                AppiumCommandExecutionHelper.PrepareArgument("seconds"));`

and xcuitest should remain as;

`Execute(AppiumDriverCommand.BackgroundApp,
                AppiumCommandExecutionHelper.PrepareArgument("seconds", AppiumCommandExecutionHelper.PrepareArgument("timeout", seconds)));`

matthew-horrocks:
Would be great to get this fixed. I'm happy to do a PR for the fix above.

akinsolb:
@matthew-horrocks go for it. I want to do a release this week, so if it doesn’t make the major release it’ll be in the minor

matthew-horrocks:
I had a look at this. Turns out it's not as simple as I first thought.

**Android**
Updated `public void BackgroundApp(int seconds)` to  `Execute(AppiumDriverCommand.BackgroundApp, AppiumCommandExecutionHelper.PrepareArgument("seconds"));` called it using `appiumDriver.BackgroundApp(30);`. This works fine for Android - the app goes into the background and returns 30 seconds later.

[...]

PR #381, opened 2019-12-13
Bugfix/309 background app

akinsolb:
## Types of changes

- [X] Bugfix (non-breaking change which fixes an issue)

## Documentation
- [X] Have you proposed a file change/ PR with appium to update documentation? 
## Integration tests
- [X] Have you provided integration tests to pass against the beta version of appium? (for Bugfix or New feature)

## Details
Updated abstract payload to match current api requirements

### DIFF
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -13,7 +13,6 @@
 //limitations under the License.
 
 using Appium.Interfaces.Generic.SearchContext;
-using Newtonsoft.Json;
 using OpenQA.Selenium.Appium.Enums;
 using OpenQA.Selenium.Appium.Interfaces;
 using OpenQA.Selenium.Appium.Service;
@@ -281,9 +280,15 @@ namespace OpenQA.Selenium.Appium
             Execute(AppiumDriverCommand.BackgroundApp,
                 AppiumCommandExecutionHelper.PrepareArgument("seconds", -1));
 
+        [Obsolete(
+            "This method is obsolete & will be removed in the next major release. Call BackgroundApp(TimeSpan) instead.")]
         public void BackgroundApp(int seconds) =>
             Execute(AppiumDriverCommand.BackgroundApp,
                 AppiumCommandExecutionHelper.PrepareArgument("seconds", seconds));
+
+        public void BackgroundApp(TimeSpan timeSpan) =>
+            Execute(AppiumDriverCommand.BackgroundApp,
+                AppiumCommandExecutionHelper.PrepareArgument("seconds", timeSpan.TotalSeconds));
 
         public AppState GetAppState(string appId) =>
             (AppState) Convert.ToInt32(Execute(AppiumDriverCommand.GetAppState,
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -279,7 +279,7 @@ namespace OpenQA.Selenium.Appium
 
         public void BackgroundApp() =>
             Execute(AppiumDriverCommand.BackgroundApp,
-                AppiumCommandExecutionHelper.PrepareArgument("seconds", null));
+                AppiumCommandExecutionHelper.PrepareArgument("seconds", -1));
 
         public void BackgroundApp(int seconds) =>
             Execute(AppiumDriverCommand.BackgroundApp,
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -283,8 +283,7 @@ namespace OpenQA.Selenium.Appium
 
         public void BackgroundApp(int seconds) =>
             Execute(AppiumDriverCommand.BackgroundApp,
-                AppiumCommandExecutionHelper.PrepareArgument("seconds",
-                    TimeSpan.FromSeconds(seconds).TotalSeconds));
+                AppiumCommandExecutionHelper.PrepareArgument("seconds", seconds));
 
         public AppState GetAppState(string appId) =>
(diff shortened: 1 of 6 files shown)


src/Appium.Net/Appium/iOS/IOSDriver.cs
```diff
@@ -227,5 +227,10 @@ public IOSDriver(AppiumLocalService service, DriverOptions driverOptions, TimeSp
         /// </summary>
         /// <returns>The image content of the clipboard as an Image object or null if there is no image on the clipboard</returns>
         public Image GetClipboardImage() => IOSCommandExecutionHelper.GetClipboardImage(this);
+
+        public new void BackgroundApp(int seconds) =>
+            Execute(AppiumDriverCommand.BackgroundApp,
+                AppiumCommandExecutionHelper.PrepareArgument("seconds",
+                    AppiumCommandExecutionHelper.PrepareArgument("timeout", TimeSpan.FromSeconds(seconds).TotalMilliseconds)));
```
mykola-mokhnach: `timeout` should not be there at all

akinsolb: Thanks, I take it this https://github.com/appium/appium-dotnet-driver/issues/309#issuecomment-459757398 is no longer relevant

KazuCocoa: Yeah, previously it was required, but it can drop for now https://github.com/appium/appium-xcuitest-driver/pull/1064
For the backword compatibility, it's better to leave TODO to remove timeout for now if you'd like to keep the timeout :)

akinsolb: Ah! thanks for that context. It's an old bug. I'll remove the timeout object to align with the api documentation

src/Appium.Net/Appium/AppiumDriver.cs
```diff
@@ -279,13 +279,12 @@ public void Rotate(Dictionary<string, int> opts)
 
         public void BackgroundApp() =>
             Execute(AppiumDriverCommand.BackgroundApp,
-                AppiumCommandExecutionHelper.PrepareArgument("seconds",
-                    AppiumCommandExecutionHelper.PrepareArgument("timeout", null)));
+                AppiumCommandExecutionHelper.PrepareArgument("seconds", null));
```
mykola-mokhnach: I suppose the default value should be `-1`, so the app is put into background forever.

akinsolb: Done

src/Appium.Net/Appium/AppiumDriver.cs
```diff
@@ -279,13 +279,12 @@ public void Rotate(Dictionary<string, int> opts)
...
             Execute(AppiumDriverCommand.BackgroundApp,
-                AppiumCommandExecutionHelper.PrepareArgument("seconds",
-                    AppiumCommandExecutionHelper.PrepareArgument("timeout", null)));
+                AppiumCommandExecutionHelper.PrepareArgument("seconds", -1));
 
         public void BackgroundApp(int seconds) =>
             Execute(AppiumDriverCommand.BackgroundApp,
                 AppiumCommandExecutionHelper.PrepareArgument("seconds",
-                    AppiumCommandExecutionHelper.PrepareArgument("timeout", seconds)));
+                    TimeSpan.FromSeconds(seconds).TotalSeconds));
```
mykola-mokhnach: I don't see any sense in converting the value from seconds to seconds. It would probably make more sense if the type of the method argument itself would be TimeSpan

test/integration/helpers/Apps.cs
```diff
@@ -63,5 +63,12 @@ public static string Get(string appKey)
             Init();
             return _testApps[appKey];
         }
+
+        public static string iosTestApp = "iosTestApp";
```
mykola-mokhnach: shouldn't they be constants?

src/Appium.Net/Appium/AppiumDriver.cs
akinsolb: I would love to change the signature to Timespan, but didn't want to break the API, especially as it'll go in a minor release.

```Timespan.FromSeconds.TotalSeconds``` just sends a double

I could just send the raw int. What do you think @mykola-mokhnach

File: test/integration/helpers/Apps.cs (hunk 4 above)
Comment (akinsolb) [reply test/integration/helpers/Apps.cs]: Can't be static and const. I'm referencing these properties in test classes instead of raw string values

File: src/Appium.Net/Appium/AppiumDriver.cs (hunk 3 above)
Comment (akinsolb) [reply src/Appium.Net/Appium/AppiumDriver.cs]: I'm tempted to break it, but since you've unified the type to seconds server side, I'll keep it as int, and remove the redundant timespan

Comment (mykola-mokhnach) [reply src/Appium.Net/Appium/AppiumDriver.cs]: The usual approach is to make this method signature deprecated and just add another one with the correct argument. Then we could remove the deprecated method in the next major release

Comment (akinsolb) [reply src/Appium.Net/Appium/AppiumDriver.cs]: Cool. I can do that

File: src/Appium.Net/Appium/AppiumDriver.cs [hunk 5]
```diff
@@ -279,13 +278,17 @@ public void Rotate(Dictionary<string, int> opts)
 
         public void BackgroundApp() =>
             Execute(AppiumDriverCommand.BackgroundApp,
-                AppiumCommandExecutionHelper.PrepareArgument("seconds",
-                    AppiumCommandExecutionHelper.PrepareArgument("timeout", null)));
+                AppiumCommandExecutionHelper.PrepareArgument("seconds", -1));
 
+        [Obsolete(
+            "This method is obsolete & will be removed in the next major release. Call BackgroundApp(TimeSpan) instead.")]
```
KazuCocoa: 👍

PR #397, opened 2020-02-19
Update AppiumOptions to work with selenium 4

LyubomirStoimchev:
## Change list
* Delete AppiumCapabilities since it inherits the deprecated DesiredCapabilities
* Use options property PlatformName instead of adding it as new option
* Make AppiumOptions to append "appium:" prefix for each option. This is needed because there is no w3c capability handling and current options are disregarded because they are not specification compliant.
* Change code using Mouse and Keyboard since they are no longer available
 
## Types of changes

## Documentation
## Integration tests

## Details

AppiumOptions use AppiumCapabilities, which uses DesiredCapabilities. In selenium version 4, DesiredCapabilities are deprecated. The changes in this PR are focused on implementing removing the deprecated dependency. A prefix "appium:" will be added to each new added option. This is needed because currently for dotnet there is no handling for options that are not w3c compliant.
I tried to implement AppiumOptions in a similar manner as the rest of the driver options(like ChromiumOptions), but I couldn't find any way to add the new options as "KnownCapabilities" that will be passed to the remote driver.


akinsolb:
@LyubomirStoimchev Can this change be done without migrating to selenium 4-beta?

LyubomirStoimchev:
@akinsolb the methods ValidateCapabilityName() and AddAdditionalOption() do not exist in 3.141.0, but they can be implemented inside AppiumOptions, instead of using them from DriverOptions.
The main issue is the interface IWritableCapabilities - it does not exist in the previous version of selenium and I cannot change it to DesiredCapabilities since it's deprecated (the access modifier has been changed from public to internal). Do you have any suggestions how to overcome this?

gravity-api:
Please try the following code.
In order to remove any dependency on deprecated interfaces I have implemented a "bridge" component, hopefully allowing me a smooth passage into Selenium 4.

I am using reflection to directly "inject" capabilities into the driver capabilities collection, bypassing all logic layer on the way (originally designed for adding non-standard capabilities) and removing the need of implementing deprecated interfaces.

This is the quick and dirty version, but you will get the idea, feel free to customize it.


```csharp
    public static class OptionsExtensions
    {
        // get capabilities interface and manipulate capabilities
        private static ICapabilities AddCapabilities(this DriverOptions driverOptions, Dictionary<string, object> rawCapabilities)
        {
            // convert options
            var options = driverOptions.ToCapabilities();

            // get capabilities field
            var isFieldNull = options
                .GetType()
                .GetField("capabilities", BindingFlags.NonPublic | BindingFlags.Instance) == null;

            Dictionary<string, object> cap = null;
            if (isFieldNull)
            {
                cap = (Dictionary<string, object>)options
                    .GetType()
                    .BaseType
                    .GetField("capabilities", BindingFlags.NonPublic | BindingFlags.Instance)
                    .GetValue(options);
            }
            else
            {
                cap = (Dictionary<string, object>)options
                    .GetType()
                    .GetField("capabilities", BindingFlags.NonPublic | BindingFlags.Instance)
                    .GetValue(options);
            }

            // exit conditions
            if (cap == null)
            {
                return options;
            }

            // add capabilities
            foreach (var item in rawCapabilities)
            {
                cap[item.Key] = item.Value;
            }
            return options;
        }
    }
```

LyubomirStoimchev:
@gravity-api Thanks for the suggestion. I tried you approach (you can see it here https://github.com/LyubomirStoimchev/appium-dotnet-driver/compare/feature/enhance-appium-options...LyubomirStoimchev:feature/update-know-caps?expand=1) but yet again I cannot think of a way to have this working on both selenium 3 and 4. The issue is that the method GenerateDesiredCapabilities() in selenium 3 returns DesiredCapabilities and in version 4 it returns IWritableCapabilities.

Another issue is that the selenium driver removes any capability that is not specification compliant (or not containing ":"), which I overcame it with using reflection, but I don't think this is the correct way. I think that appending "appium:" before the capability is the correct way, since for chrome it's done in a similar manner.

mykola-mokhnach:
Unfortunately we cannot put selenium 4 alpha integration into `master`, since most of the users rely on the stable implementation and they don't want the `WIP` stuff to be there.

Consider making a fork for now

gravity-api:
@LyubomirStoimchev hi :)
The solution I gave you solves exactly that. It will inject capabilities no matter how they are exposed on the front end of Selenium.

It does not matter if it is exposed as DesiredCapabilites or IWriteableCapabilites or if Selenium reject any capabilities.

The code I have sent you will directly put it in the capabilities sent to to driver (BYPASSING all selenium restrictions). Please see the following code for with implementation sample.

```csharp
using OpenQA.Selenium;
using OpenQA.Selenium.Chrome;
using System.Collections.Generic;
using System.Reflection;
using System.Threading;

namespace CapabilitesSample
{
    internal static class Program
    {
        private static void Main()
        {
            // create options
            var options = new ChromeOptions();

            // add whatever capabilities into capabilities dictionary
            options.AddCapabilities(new Dictionary<string, object>
            {
                ["cap1"] = true,
                ["cap2"] = 1,
                ["cap3"] = "text",
                ["cap4"] = new { Prop1 = "1", Prop2 = 2 }
            });

            // create new driver
            var driver = new ChromeDriver(options);
            Thread.Sleep(3000);
            driver.Dispose();
        }
    }

    public static class OptionsExtensions
    {
        // get capabilities interface and manipulate capabilities
        public static ICapabilities AddCapabilities(this DriverOptions driverOptions, Dictionary<string, object> rawCapabilities)
        {
            // convert options
            var options = driverOptions.ToCapabilities();

            // get capabilities field
            var isFieldNull = options
                .GetType()
                .GetField("capabilities", BindingFlags.NonPublic | BindingFlags.Instance) == null;

            Dictionary<string, object> cap = null;
            if (isFieldNull)
            {
                cap = (Dictionary<string, object>)options
                    .GetType()
                    .BaseType
                    .GetField("capabilities", BindingFlags.NonPublic | BindingFlags.Instance)
                    .GetValue(options);
            }
            else
            {
                cap = (Dictionary<string, object>)options
                    .GetType()
                    .GetField("capabilities", BindingFlags.NonPublic | BindingFlags.Instance)
                    .GetValue(options);
            }

            // exit conditions
            if (cap == null)
            {
                return options;
            }

            // add capabilities
            foreach (var item in rawCapabilities)
            {
                cap[item.Key] = item.Value;
            }
            return options;
        }
    }
}
```

LyubomirStoimchev:
@gravity-api I tried the code with creating the remote driver but the capabilities were removed (https://github.com/SeleniumHQ/selenium/blob/ff7f4b285211e11c5099b1a25ab3aa8807f4dc8b/dotnet/src/webdriver/Remote/RemoteWebDriver.cs#L960) because they were not known capabilities (I later injected with reflection new values to the known capabilities in CapabilityType class) but the whole point of the pull request is to enable appium to use selenium 4 version.
Anyway, I have already forked and made changes so that it runs with selenium 4(which unfortunately I have to use) and I thought why not create a PR here so that someone might find this useful.

@mykola-mokhnach I agree with you that an alpha version should not be put in master. I hope that in the future when selenium 4 is officially released the changes in the PR might be helpful to some extent.

gravity-api:
@LyubomirStoimchev Hi
Thank you for the detailed explanation - I think I was not clear enough.
The point of my code was to avoid implementing the deprecated interfaces for creating custom capabilities which passed to the driver.

The "remove" of capabilities you are talking about is the normal behavior - when you are working locally, but if you want to work with Browser Stack, Lambda Testing or Sauces Lab it will not remove the capabilities each of the platform uses - the AddAdditional/Know capabilities method will not work for these  platform capabilities.

That being said, you can override the actual request for session creating and send it raw - at that point you completely change the way you get session (I use this method for some old browser testing)

Please let me know if you want code sample

PR #410, opened 2020-06-05
feat: Added builder pattern for creating valid ByAndroidUIAutomator statements

sparerd:
## Change list

- Added 2 new fluent-builder style classes (`AndroidUiSelector`, `AndroidUiScrollable`) for building valid `UiSelector` and `UiScrollable` calls.
    - Added new interface `IUiAutomatorStatementBuilder` which both new classes implement
- Added method overloads to `IFindByAndroidUIAutomator`
    - `FindElementByAndroidUIAutomator(IUiAutomatorStatementBuilder)`
    - `FindElementsByAndroidUIAutomator(IUiAutomatorStatementBuilder)`
- Added ctor overload to `ByAndroidUIAutomator`
    - `ByAndroidUIAutomator(IUiAutomatorStatementBuilder selector)`
- Added static `By.AndroidUIAutomator(IUiAutomatorStatementBuilder)` overload
 
## Types of changes

## Documentation

I have not submitted a doc change as there doesn't seem to currently be a good place to document this change.

## Integration tests

## Details

While it is nice to be able to use UiAutomator directly, it is a bit of a pain to build using strings directly. This PR adds two fluent-builder style classes to guide the library user through creating valid UiAutomator statements. Each builder can be compiled to a string and used with existing methods, though I've added several interface/ctor overloads to make this more discoverable.

If it is desirable to keep the `IFindByAndroidUIAutomator` interface more simple, the overloads I've added can be converted to extension methods.

Each builder has an `AddRawText(string)` method which will add any provided text directly to the builder. This is to allow extensibility by library users in the case that the method they require is not included in the fluent builder wrapper. At this time, UiSelector is fully wrapped, but UiScrollable is pretty bare-bones.

**Ex. 1: Find an element by UiSelector**
```csharp
var options = new AppiumOptions {PlatformName = "Android"};
var driver = new AndroidDriver<AndroidElement>(options);

var element = driver.FindElementByAndroidUIAutomator(new AndroidUiSelector()
    .ClassNameEquals("android.widget.TextView")
    .Instance(1));
```

**Ex. 2: Scroll to an element which matches a selector**
```csharp
var options = new AppiumOptions {PlatformName = "Android"};
var driver = new AndroidDriver<AndroidElement>(options);

var element = driver.FindElementByAndroidUIAutomator(new AndroidUiScrollable()
    .ScrollIntoView(new AndroidUiSelector()
        .ResourceIdMatches(".*my_unique_id")
        .TextEquals("Test Element")));
```

**Ex. 3: Scroll to an element which matches a selector, using preexisting overload which is not aware of the builder**
```csharp
var options = new AppiumOptions {PlatformName = "Android"};
var driver = new AndroidDriver<AndroidElement>(options);

var element = driver.FindElementByAndroidUIAutomator(new AndroidUiScrollable()
    .ScrollIntoView(new AndroidUiSelector()
        .ResourceIdMatches(".*my_unique_id")
        .TextEquals("Test Element"))
    .Compile());
```

**Other supported uses**
```csharp
var staticBy = MobileBy.AndroidUIAutomator(new AndroidUiSelector().ResourceIdMatches(".*my_unique_id"));

var newBy = new ByAndroidUIAutomator(new AndroidUiSelector().ResourceIdMatches(".*my_unique_id"))
```

### DIFF
--- a/src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
+++ b/src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
@@ -1,0 +1,92 @@
+﻿using System.Text;
+using OpenQA.Selenium.Appium.Android.Enums;
+using OpenQA.Selenium.Appium.Interfaces;
+
+namespace OpenQA.Selenium.Appium.Android.UiAutomator
+{
+    /// <summary>
+    /// 
+    /// </summary>
+    /// <remarks>
+    /// Docs: https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable
+    /// </remarks>
+    public class AndroidUiScrollable : IUiAutomatorStatementBuilder
+    {
+        private readonly StringBuilder _builder;
+
+        /// <summary>
+        /// Creates a new scrollable searcher which will match the first scrollable widget
+        /// in the view.
+        /// </summary>
+        public AndroidUiScrollable()
+            : this(new AndroidUiSelector().IsScrollable(true))
+        {
+        }
+
+        /// <summary>
+        /// Creates a new scrollable searcher which will match the first widget
+        /// which matches the given UISelector.
+        /// </summary>
+        /// <param name="uiSelector"></param>
+        public AndroidUiScrollable(AndroidUiSelector uiSelector)
+        {
+            _builder = new StringBuilder().AppendFormat("new UiScrollable({0})", uiSelector.Compile());
+        }
+
+        /// <summary>
+        /// Sets the scrolling direction of the list.
+        /// </summary>
+        /// <param name="direction"></param>
+        /// <returns></returns>
+        public AndroidUiScrollable SetScrollDirection(ListDirection direction)
+        {
+            _builder.AppendFormat(".setAs{0}List()", direction);
+            return this;
+        }
+
+        /// <summary>
+        /// Perform a scroll forward action to move through the scrollable layout element until a visible item that matches the selector is found.
+        /// Maps to the UiScrollable.scrollIntoView(UiSelector) method.
+        /// </summary>
+        /// <param name="uiSelector"></param>
+        /// <remarks>https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable#scrollintoview</remarks>
+        public AndroidUiScrollable ScrollIntoView(AndroidUiSelector uiSelector)
+        {
+            _builder.AppendFormat(".scrollIntoView({0})", uiSelector.Compile());
+            return this;
+        }
+
+        /// <summary>
+        /// Append raw text to this <see cref="AndroidUiScrollable"/> instance. The target language is Java.
+        /// Text entered here will not be checked for validity. Use this at your own risk.
+        /// </summary>
+        /// <param name="text">
+        /// Text to be appended to the UiScrollable command builder.
+        /// </param>
+        public AndroidUiScrollable AddRawText(string text)
+        {
+            _builder.Append(text);
+            return this;
+        }
+
+        public string Compile()
+        {
+            return Compile(false);
+        }
+
+        /// <summary>
+        /// Compiles the current UiScrollable statements into a valid Java string which can be executed.
+        /// </summary>
+        /// <param name="terminateStatement">
+        /// Should the statement be returned with a semicolon terminator at the end. Defaults to false.
+        /// The terminator is only appended to the returned statement string - this <see cref="AndroidUiScrollable"/> is not affected.
+        /// </param>
+        public string Compile(bool terminateStatement)
+        {
+            if (terminateStatement)
+                return _builder + ";";
+
+            return _builder.ToString();
+        }
+    }
+}
--- a/src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
+++ b/src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
@@ -19,7 +19,7 @@ using OpenQA.Selenium.Appium.Interfaces;
 namespace OpenQA.Selenium.Appium.Android.UiAutomator
 {
     /// <summary>
-    /// 
+    /// A convenience class to wrap Android UiScrollable method calls.
     /// </summary>
     /// <remarks>
     /// Docs: https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable
--- a/src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
+++ b/src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
@@ -12,6 +12,7 @@
 //See the License for the specific language governing permissions and
 //limitations under the License.
 
+using System;
 using System.Text;
 using OpenQA.Selenium.Appium.Android.Enums;
 using OpenQA.Selenium.Appium.Interfaces;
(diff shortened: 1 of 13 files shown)


src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
```diff
@@ -0,0 +1,106 @@
...
+            return this;
+        }
+
+        /// <summary>
+        /// Perform a scroll forward action to move through the scrollable layout element until a visible item that matches the selector is found.
+        /// Maps to the UiScrollable.scrollIntoView(UiSelector) method.
+        /// </summary>
+        /// <param name="uiSelector"></param>
+        /// <remarks>https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable#scrollintoview</remarks>
+        public AndroidUiScrollable ScrollIntoView(AndroidUiSelector uiSelector)
```
mykola-mokhnach, line 67: UiScrollable supports much more than this: https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable

test/integration/Android/ElementTest.cs
```diff
@@ -1,8 +1,10 @@
 ﻿using Appium.Net.Integration.Tests.helpers;
```
mykola-mokhnach, line 1: I would be happy to see more unit tests, which cover most of the builder methods. This stuff could easily be tested in isolation.

sparerd: Will do. Is it enough to validate that the correct string is built, or would you prefer a full integration test where the built statement is used to find an element?

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
sparerd: I can add the rest

sparerd:
Renamed `Compile()` to `Build()` on the builder interface as requested.

test/integration/Android/ElementTest.cs
mykola-mokhnach: the first one is pretty enough for a unit test

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
mykola-mokhnach: Some methods there, that don't result in UiScrollable, might fail in Appium below 1.18.0. Since v. 1.18.0 they simply result in the most recent scrollable element in the chain

test/integration/Android/ElementTest.cs
sparerd: Unit tests added. Feels a little weird testing strings, but it did uncover one incorrect builder statement. Let me know if more needs to be done.

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
```diff
@@ -41,14 +42,316 @@ public AndroidUiScrollable()
...
+        /// Content-description to find in the children of the childPattern match (may be a partial match)
+        /// </param>
+        /// <param name="allowScrollSearch">
+        /// Set to true if scrolling is allowed
+        /// </param>
+        /// <remarks>https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable#getchildbydescription_1</remarks>
+        public TerminatedStatementBuilder GetChildByDescription(AndroidUiSelector uiSelector, string description, bool allowScrollSearch = true)
+        {
+            if (uiSelector == null)
+                throw new ArgumentNullException(nameof(uiSelector));
```
mykola-mokhnach, line 144: could the https://stackoverflow.com/questions/29184887/best-way-to-check-for-null-parameters-guard-clauses syntax be used instead to make the code shorter?

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
sparerd: Most UiScrollable methods are now included.

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
```diff
@@ -41,14 +42,316 @@ public AndroidUiScrollable()
...
+        /// <param name="uiSelector"></param>
+        /// <param name="instance"></param>
+        /// <remarks>https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable#getchildbyinstance</remarks>
+        public TerminatedStatementBuilder GetChildByInstance(AndroidUiSelector uiSelector, int instance)
+        {
+            if (uiSelector == null)
+                throw new ArgumentNullException(nameof(uiSelector));
+
+            if (instance < 0)
+                throw new ArgumentOutOfRangeException(nameof(instance), "Must be 0 (zero) or greater");
```
mykola-mokhnach, line 171: this check could be extracted to a separate helper method, for example `requirePositiveInteger`

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
mykola-mokhnach: In general, I like the Kotlin's approach to it: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/require.html

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
sparerd: Since the non-negative guards exist in both new classes, should I extract the guard check to a new internal class? Static?

test/integration/Android/AndroidUiScrollableTests.cs
```diff
@@ -0,0 +1,193 @@
+﻿using System;
+using NUnit.Framework;
+using OpenQA.Selenium.Appium.Android.Enums;
+using OpenQA.Selenium.Appium.Android.UiAutomator;
+
+namespace Appium.Net.Integration.Tests.Android
```
mykola-mokhnach, line 6: 👍

test/integration/Android/AndroidUiSelectorTests.cs
```diff
@@ -0,0 +1,232 @@
...
+        public void AddRawTextAppendsText()
+        {
+            var statment = _sut.AddRawText("@").Build();
+            Assert.AreEqual(CtorStatement + "@", statment);
+        }
+
+        [Test]
+        public void CanChainCalls()
+        {
+            var statment = _sut
```
mykola-mokhnach, line 223: statement

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
mykola-mokhnach: it's up to you, I'm not an expert here. The main purpose is to avoid unnecessary code duplication

test/integration/Android/ElementTest.cs
mykola-mokhnach: > Feels a little weird testing strings, but it did uncover one incorrect builder statement

And this is great

test/integration/Android/AndroidUiSelectorTests.cs
sparerd: Fixed typo

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
```diff
@@ -190,11 +184,8 @@ public TerminatedStatementBuilder GetChildByInstance(AndroidUiSelector uiSelecto
         public TerminatedStatementBuilder GetChildByText(AndroidUiSelector uiSelector, string text,
             bool allowScrollSearch = true)
         {
-            if (uiSelector == null)
-                throw new ArgumentNullException(nameof(uiSelector));
-
-            if (text == null)
-                throw new ArgumentNullException(nameof(text));
+            _ = uiSelector ?? throw new ArgumentNullException(nameof(uiSelector));
```
mykola-mokhnach, line 187: this could be further simplified to a method, which looks like

```
object<T> requireNonNull(T o) {
  return o ?? throw new ArgumentNullException(nameof(o))
} 

```

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
sparerd: A project I work on uses the extension method approach for their null guards, has a similar effect.

Example of the way I was going to handle the non-neg guard from you other comment:
```csharp
public void aaaa(int value)
{
    value.EnsureNumberIsPositive(nameof(value));
    // etc etc etc...
}
```

If you like this approach, I can do the null guards the same way. The one downside is that the top of the stack trace will have the guard method rather than the actual method the user called. Though it's pretty minor since the stack will at least still have the user-called method on it.

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
sparerd: Sure, I convert the null guards to a throw-or-access style

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
mykola-mokhnach: I simply see an unnecessary code duplication, which could be easily eliminated.  
It would be also effective to return the same value from the verification method if the verification passes. This way we could also save one more line of code

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
sparerd: Take a look at the new iteration on guard clauses and let me know what you think

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
```diff
@@ -0,0 +1,402 @@
...
+        /// <param name="description">
+        /// Content-description to find in the children of the childPattern match (may be a partial match)
+        /// </param>
+        /// <param name="allowScrollSearch">
+        /// Set to true if scrolling is allowed
+        /// </param>
+        /// <remarks>https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable#getchildbydescription_1</remarks>
+        public TerminatedStatementBuilder GetChildByDescription(AndroidUiSelector uiSelector, string description, bool allowScrollSearch = true)
+        {
+            uiSelector.RequireNotNull(nameof(uiSelector));
```
mykola-mokhnach, line 136: This is a good idea to make it an extension method. I like it. Btw, do we need this argument at all. Isn't it possible to to do something like `nameof(this)` inside the extension method?

Also, like I said, we could save one line of code by calling the verification directly when the corresponding var is used:

```
var forkedBuilder = new StringBuilder(_builder.ToString());
             forkedBuilder.AppendFormat(".getChildByDescription({0}, \"{1}\", {2})", 
                 uiSelector.RequireNotNull().Build(), 
                 description.RequireNotNull(),
                 allowScrollSearch.ToString().ToLowerInvariant());
             return new TerminatedStatementBuilder(forkedBuilder);

```

mykola-mokhnach: one more concern: `uiSelector.RequireNotNull(nameof(uiSelector));` is going to throw NPE without calling `RequireNotNull` if `uiSelector` is null. That is why exactly this check cannot be an extension, but a separate function

sparerd: > Btw, do we need this argument at all. Isn't it possible to to do something like nameof(this) inside the extension method?

Unfortunately, it is required. There is no native way to automatically pass along the *real* name of the argument. Putting the `nameof()` in the extension method would always report the same variable name (in our case, "value"). This leads to quite a bit of confusion when a single method has several nullable arguments. Example in code:

```csharp
public static T RequireNotNull<T>(T value)
{
    if (value != null) return;
    throw new ArgumentNullException(nameof(value));
}

// The if either argument is null, the exception will report
// that the variable named "value" is null, leading to some confusion
// over which arg was actually null
public void aaaaa(AndroidUiSelector selector, string name)
{
    selector.RequireNotNull();
    name.RequireNotNull();
}
```

> Also, like I said, we could save one line of code by calling the verification directly when the corresponding var is used:

I decided to keep the guard clauses up front since otherwise we will always do the work of cloning the string builder, regardless of whether we *should* do the work. If a guard clause fails, we have to abandon that cloning work, wasting cycles. Admittedly, it usually won't be that much work that gets wasted (fairly small strings) and ideally wouldn't happen often. With that said, if you would still like those cases in-lined, let me know and I'll make the change.

> one more concern: uiSelector.RequireNotNull(nameof(uiSelector)); is going to throw NPE without calling RequireNotNull if uiSelector is null. That is why exactly this check cannot be an extension, but a separate function

In C#, extension functions compile to static method calls. It will still be safe to call RequireNotNull on a null value.

An example call, shown in CIL:
`IL_0007: call string OpenQA.Selenium.Appium.GuardClauses::RequireNotNull<string>(!!0,  string)`

mykola-mokhnach: > If a guard clause fails, we have to abandon that cloning work, wasting cycles. 

I wouldn't worry about that. Let the interpreter to do micro-optimizations. 

> In C#, extension functions compile to static method calls. It will still be safe to call RequireNotNull on a null value.

Oh, magic, magic everywhere.

sparerd: In-lined final set of guard clauses.

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiScrollable.cs
```diff
@@ -0,0 +1,383 @@
...
+        /// Performs a forward fling with the default number of fling steps (5). If the swipe direction is set
+        /// to vertical, then the swipes will be performed from bottom to top. If the swipe direction is set to
+        /// horizontal, then the swipes will be performed from right to left. Make sure to take into account
+        /// devices configured with right-to-left languages like Arabic and Hebrew.
+        /// Maps to the UiScrollable.flingForward() method.
+        /// </summary>
+        /// <remarks>https://developer.android.com/reference/android/support/test/uiautomator/UiScrollable#flingforward</remarks>
+        public TerminatedStatementBuilder FlingForward()
+        {
+            var forkedBuilder = new StringBuilder(_builder.ToString());
```
mykola-mokhnach, line 78: `var forkedBuilder = new StringBuilder(_builder.ToString());`

could also be extracted into a separate method, for example:
```
private forkBuilder ()
{
  return new StringBuilder(_builder.ToString());
}


private append (String s)
{
   return new TerminatedStatementBuilder(forkBuilder().AppendFormat(s));
}

public TerminatedStatementBuilder FlingForward()
{
   return append(".flingForward()");
}
....
```

mykola-mokhnach:
@sparerd Do you have time to finish the PR? We'd be happy to merge it when it's done

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiSelector.cs
```diff
@@ -0,0 +1,477 @@
...
+            return this;
+        }
+
+        /// <summary>
+        /// Set the search criteria to match the package name of the application that contains the widget.
+        /// Maps to the UiSelector.packageNameMatches(String) method.
+        /// </summary>
+        /// <param name="regex"></param>
+        /// <remarks>https://developer.android.com/reference/androidx/test/uiautomator/UiSelector#packagenamematches</remarks>
+        public AndroidUiSelector PackageNameMatches(string regex)
```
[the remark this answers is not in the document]
sparerd: I'm not sure how using the Regex class would add safety here. Do you have a specific scenario in mind?

Other pros/cons for using the Regex class:
- Pro: it would add a bit of explicitness to the method signature if someone doesn't read the argument name
- Pro: you would get regex syntax highlighting when using the Regex ctor
- Con: since we don't actually use the Regex class for the matching, some options on the Regex constructor would be misleading for an end user. Example: in the following call, the RegexOption and timeout arguments would be completely ignored `uiSelector.PackageNameMatches(new Regex(@"myapp$", RegexOptions.ECMAScript, matchTimeout:TimeSpan.FromSeconds(3)))`

I lean toward keeping it a string-typed argument - I think it causes less confusion.

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiSelector.cs
```diff
@@ -0,0 +1,477 @@
...
+        /// <summary>
+        /// Set the search criteria to match the content-description property for a widget. The content-description
+        /// is typically used by the Android Accessibility framework to provide an audio prompt for the widget when
+        /// the widget is selected. The content-description for the widget must match exactly with the string in your
+        /// input argument. Matching is case-sensitive.
+        /// Maps to the UiSelector.description(String) method.
+        /// </summary>
+        /// <param name="description"></param>
+        /// <remarks>https://developer.android.com/reference/androidx/test/uiautomator/UiSelector#description</remarks>
+        public AndroidUiSelector DescriptionEquals(string description)
```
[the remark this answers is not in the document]
sparerd: I'm not sure if we need to do anything special for line breaks, I'll write a quick test for that. We will probably need to escape any double quotes though.

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiSelector.cs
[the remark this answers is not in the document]
mykola-mokhnach: I'm fine with the current implementation.

src/Appium.Net/Appium/Android/UiAutomator/AndroidUiSelector.cs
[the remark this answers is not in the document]
mykola-mokhnach: yes, unit tests would be nice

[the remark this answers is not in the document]
sparerd: See latest 2 commits for integration test using line breaks - works as expected. Unexpectedly, double quotes also work without any extra escaping required.

PR #417, opened 2020-08-28
feat: Provide way to pass arguments to NodeJS

sparerd:
## Change list

Added option to provide command line arguments to Node in the `AppiumServiceBuilder`.
 
## Types of changes

## Documentation
## Integration tests

## Details
I've added this feature to support loading Appium from yarn PnP-ified packages (https://next.yarnpkg.com/features/pnp). We needed to pass `--require .pnp.js` to Node in order to include support for reading node packages from zip files. This allows us to reduce build time and have more manageable Node artifacts (~100 zip files rather than 1000s of tiny js files).

### DIFF
--- a/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
+++ b/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
@@ -496,7 +496,7 @@ namespace OpenQA.Selenium.Appium.Service
 
                 if (ServerOptions != null)
                 {
-                    argList.AddRange(ServerOptions.Argiments);
+                    argList.AddRange(ServerOptions.Arguments);
                 }
 
                 string result = string.Empty;
--- a/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
+++ b/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
@@ -40,7 +40,7 @@ namespace OpenQA.Selenium.Appium.Service
         private FileInfo NodeJS;
         private IDictionary<string, string> EnvironmentForAProcess;
         private string PathToLogFile;
-        private IDictionary<string, string> NodeOptions;
+        private string[] NodeOptions = new string[0];
 
 
         private static Process StartSearchingProcess(string file, string arguments)
@@ -311,22 +311,22 @@ namespace OpenQA.Selenium.Appium.Service
         /// Command line arguments that will be passed to NodeJS when it starts up.
         /// </summary>
         /// <param name="arguments">
-        /// A collection of argument name and argument value pairs that will be
-        /// passed to NodeJS.
+        /// A collection of arguments that will be passed to NodeJS. Spaces will automatically
+        /// be added between arguments. Arguments cannot be null, empty strings, or only whitespace.
         /// </param>
         /// <returns></returns>
-        public AppiumServiceBuilder WithNodeArguments(IDictionary<string, string> arguments)
+        public AppiumServiceBuilder WithNodeArguments(params string[] arguments)
         {
             if (arguments == null)
             {
                 throw new ArgumentNullException(nameof(arguments), "Argument cannot be null");
             }
 
-            foreach (var key in arguments.Keys)
+            foreach (var arg in arguments)
             {
-                if (string.IsNullOrEmpty(key))
+                if (string.IsNullOrWhiteSpace(arg))
                 {
-                    throw new ArgumentException("Node arguments cannot have null or empty keys", nameof(arguments));
+                    throw new ArgumentException("Node arguments cannot be null, empty, or only whitespace.", nameof(arguments));
                 }
             }
 
@@ -479,17 +479,7 @@ namespace OpenQA.Selenium.Appium.Service
                 List<string> argList = new List<string>();
                 CheckAppiumJS();
 
-                if (NodeOptions != null)
-                {
-                    foreach (var argPair in NodeOptions)
-                    {
-                        argList.Add(argPair.Key);
-                        if (!string.IsNullOrWhiteSpace(argPair.Value))
-                        {
-                            argList.Add(argPair.Value);
-                        }
-                    }
-                }
+                argList.AddRange(NodeOptions);
 
                 argList.Add($"\"{AppiumJS.FullName}\"");
                 argList.Add("--port");
--- a/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
+++ b/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
@@ -40,6 +40,7 @@ namespace OpenQA.Selenium.Appium.Service
         private FileInfo NodeJS;
         private IDictionary<string, string> EnvironmentForAProcess;
         private string PathToLogFile;
+        private IDictionary<string, string> NodeOptions;
 
 
         private static Process StartSearchingProcess(string file, string arguments)
@@ -306,6 +307,20 @@ namespace OpenQA.Selenium.Appium.Service
             return this;
         }
 
+        /// <summary>
+        /// Command line arguments to pass to NodeJS when it starts up.
+        /// </summary>
+        /// <param name="arguments">
+        /// A collection of argument name and argument value pairs that will be
+        /// passed to NodeJS.
+        /// </param>
+        /// <returns></returns>
+        public AppiumServiceBuilder WithNodeArguments(IDictionary<string, string> arguments)
+        {
+            NodeOptions = arguments;
+            return this;
+        }
+
         private void CheckAppiumJS()
         {
             if (AppiumJS != null)
@@ -450,6 +465,16 @@ namespace OpenQA.Selenium.Appium.Service
             {
                 List<string> argList = new List<string>();
                 CheckAppiumJS();
+
+                if (NodeOptions != null)
+                {
+                    foreach (var argPair in NodeOptions)
+                    {
+                        argList.Add(argPair.Key ?? string.Empty);
+                        argList.Add(argPair.Value ?? string.Empty);
+                    }
+                }
+
                 argList.Add($"\"{AppiumJS.FullName}\"");
                 argList.Add("--port");
                 argList.Add($"\"{Port}\"");
--- a/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
+++ b/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
@@ -308,7 +308,7 @@ namespace OpenQA.Selenium.Appium.Service
         }
 
         /// <summary>
-        /// Command line arguments to pass to NodeJS when it starts up.
+        /// Command line arguments that will be passed to NodeJS when it starts up.
         /// </summary>
         /// <param name="arguments">
         /// A collection of argument name and argument value pairs that will be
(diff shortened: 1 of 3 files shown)


akinsolb (approved):

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
```diff
@@ -450,6 +478,19 @@ private string Args
             {
                 List<string> argList = new List<string>();
                 CheckAppiumJS();
+
+                if (NodeOptions != null)
+                {
+                    foreach (var argPair in NodeOptions)
+                    {
+                        argList.Add(argPair.Key);
+                        if (!string.IsNullOrEmpty(argPair.Value))
```
mykola-mokhnach, line 487: null verification is already done in the builder

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
```diff
@@ -306,6 +307,33 @@ public AppiumServiceBuilder WithStartUpTimeOut(TimeSpan startUpTimeout)
...
 
+        /// <summary>
+        /// Command line arguments that will be passed to NodeJS when it starts up.
+        /// </summary>
+        /// <param name="arguments">
+        /// A collection of argument name and argument value pairs that will be
+        /// passed to NodeJS.
+        /// </param>
+        /// <returns></returns>
+        public AppiumServiceBuilder WithNodeArguments(IDictionary<string, string> arguments)
```
mykola-mokhnach, line 318: arguments are not necessarily pairs, there also might be single switchers

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
```diff
@@ -450,6 +478,19 @@ private string Args
...
                 CheckAppiumJS();
+
+                if (NodeOptions != null)
+                {
+                    foreach (var argPair in NodeOptions)
+                    {
+                        argList.Add(argPair.Key);
+                        if (!string.IsNullOrEmpty(argPair.Value))
+                        {
+                            argList.Add(argPair.Value);
```
mykola-mokhnach, line 489: should there be a check if value has spaces?

test/integration/ServerTests/AppiumLocalServerLaunchingTest.cs
```diff
@@ -287,5 +289,53 @@ public void CheckTheAbilityToDefineTheDesiredLogFile()
...
+
+        [Test]
+        public void CheckAbilityToSetNodeArguments()
+        {
+            var service = new AppiumServiceBuilder()
+                .WithStartUpTimeOut(TimeSpan.FromMilliseconds(500)) // we expect the Appium startup to fail, so fail quickly
+                .WithNodeArguments(new Dictionary<string, string>
+                {
+                    // show Node version and exit
+                    {"--version", string.Empty}
```
mykola-mokhnach, line 301: I'd rather provide arguments as array

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
sparerd: Check is still necessary since field isn't set on class creation. If you are ok with always initializing `NodeOptions` then we can ditch this null check.

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
sparerd: This is the same pattern used in `OptionsCollector.AddArguments` and `AppiumServiceBuilder.WithEnvironment`. If you prefer a list I can change it. Not much difference either way.

mykola-mokhnach: for env dictionary makes sense, because there all entries are naturally key-value pairs, which is not the case for command line options

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
sparerd: Fixed

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
sparerd: Unnecessary since last commit

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
sparerd: Changed in latest

test/integration/ServerTests/AppiumLocalServerLaunchingTest.cs
sparerd: Changed in latest

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
```diff
@@ -306,6 +307,33 @@ public AppiumServiceBuilder WithStartUpTimeOut(TimeSpan startUpTimeout)
...
+            if (arguments == null)
+            {
+                throw new ArgumentNullException(nameof(arguments), "Argument cannot be null");
+            }
+
+            foreach (var arg in arguments)
+            {
+                if (string.IsNullOrWhiteSpace(arg))
+                {
+                    throw new ArgumentException("Node arguments cannot be null, empty, or only whitespace.", nameof(arguments));
```
mykola-mokhnach, line 329: I would rather show the index of the failing argument here to help users to fix the issue

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
```diff
@@ -450,6 +478,9 @@ private string Args
             {
                 List<string> argList = new List<string>();
                 CheckAppiumJS();
+
+                argList.AddRange(NodeOptions);
```
mykola-mokhnach, line 482: again, I'm not quite sure about escaping arguments having spaces. Looking at the code below this looks mandatory. Although, I would rather expect this behaviour to be automatic when we pass arguments array to the downstream command execution function.

I did not investigate the reason `Args` property is a string rather than a list, but this seems rather suspicious/fragile to me. @akinsolb what do you think?

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
```diff
@@ -450,6 +478,9 @@ private string Args
...
+
                 argList.Add($"\"{AppiumJS.FullName}\"");
                 argList.Add("--port");
                 argList.Add($"\"{Port}\"");
```
mykola-mokhnach, line 486: nit: ServerOptions.ArgIments

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
sparerd: Added

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
akinsolb: I agree. We should probably create a private method that formats/escapes arguments in this class

mykola-mokhnach: should they be escaped in general? usually all exec-like function allow to pass arrays and know how to escape these without doing the routine on the client side. Maybe you know if the downstream function here supports that, so we could get rid of string representation and "manual" escaping at all?

akinsolb: There's no magic like that here AFIK. They need to be escaped. The only reason why the manual ones exist below is because a bug was raised for not handling it.

mykola-mokhnach: it looks like the standard Process class in sharp is less intellectual that I thought:
- http://csharptest.net/529/how-to-correctly-escape-command-line-arguments-in-c/index.html
- https://stackoverflow.com/questions/5510343/escape-command-line-arguments-in-c-sharp

Although, it always makes sense to use verified solutions rather than invent own wheels

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
sparerd: Fixed

src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs
mykola-mokhnach: @sparerd Could you please add a separate method for arguments quoting (for example from the links above) and then add unit tests to verify that arguments containing spaces and/or quotes are escaped properly?

sparerd: Because this new method takes in arbitrary arguments that will be passed along, it seems unnecessary and perhaps even dangerous to try to escape them. If we miss an edge case and cause a bug, it could prevent valid inputs. In that case the user is stuck - they would need to wait for a bug fix or try to find a work around.

I've been in that situation before, and fighting with Windows command line escaping *and* a library's attempt to make escaping easier isn't fun. You then need to write the correctly escaped string, then modify it further to avoid the library escaping code. As a user, I would much rather not have that feature. Or at least have a way to disable it.

It makes a bit more sense having escaping built in for the other builder methods like `WithLogFile` or the port selection methods because you do not have direct control other the string representation of those arguments. With `WithNodeArguments` users must necessarily have full control over the input. This could be made more explicit by renaming to `WithRawNodeArguments` or something like that and add some notes in the method docs.

mykola-mokhnach: this makes a lot of sense. 

The reason I touched this topic is that there was an issue is java client while passing complicated json-string (this is a very common and valid case if one wants to have default capabilities). This type of data is quite complicated and must be escaped properly. Usually manual escaping is quite a tricky task and unfortunately the most of the users are unable to solve it. So I would rather stick to an option where we have this behaviour under a flag. Let say by default the library tries to escape args, but in case your input data is too complicated then there is a possibility to provide a flag/use an alternative method to not escape custom args. What do you think about such approach?

sparerd: I can understand the desire to reduce friction and user support requests, but I'm very reluctant to add a feature (the auto-escaping) which I myself will avoid using, especially when providing args to Node is probably already a pretty niche capability. I'm worried that the escaping development will take more effort than I'm willing to put in.

At the time of this PR, most Node args seem to be flags or require small and well defined inputs. A few take paths, which are easy enough to handle. The only really scary ones (`--evaluate`) could fall pretty easily in the "dont do that" category and have other options which are easier to escape and maintain (for evaluate, just put the code in a JS file to execute).

mykola-mokhnach: ok, then it would only be necessary to mention that in the method's docstring, that the args must be escaped manually if needed, otherwise I'm fine with the PR. Thanks for your work.

akinsolb: Happy to merge @sparerd  when the xml comments provides notice to escape strings

sparerd: I've updated the arg comment - let me know if anything else should be added.

akinsolb:
I appreciate your contribution @sparerd !

PR #422, opened 2020-09-03
feat: Get Performance Data & Performance Data Types

### LINKED ISSUE #361: Add Get Performance Data endpoint
## Description
Implement Get Performance Data endpoint to return the information of the system state which is supported to read as like cpu, memory, network traffic, and battery.

## Endpoint Documentation
Endpoint documentation: http://appium.io/docs/en/commands/device/performance-data/get-performance-data/

akinsolb:
closes #361

## Change list

Please provide briefly described change list which are you going to propose. 
 
## Types of changes

## Documentation
- [x] Have you proposed a file change/ PR with appium to update documentation? 
## Integration tests

## Details
Adds support for getting performance data and performance data types on Android
Adds the following methods to AndroidDriver.cs

GetPerformanceDataTypes()
GetPerformanceData(packageName,dataType,dataReadTimeout)


akinsolb:
I couldn't get 'cpuinfo' to work/return a good response. An error occurred each time, so I commented out the assertion

mykola-mokhnach:
@akinsolb Please provide the actual server error, perhaps I could fix it

src/Appium.Net/Appium/Android/AndroidDriver.cs
```diff
@@ -213,7 +214,21 @@ public ConnectionType ConnectionType
 
         protected override RemoteWebElementFactory CreateElementFactory() => new AndroidElementFactory(this);
 
-        #region locking
+        #region Device Performance Data
+
+        public IList<object> GetSupportPerformanceData(string packageName, string dataType,
```
mykola-mokhnach, line 219: What the `Support` prefix means here?

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace OpenQA.Selenium.Appium.Interfaces
+{
+    public interface IHasSupportedPerformanceData
+    {
+        /// <summary>
+        /// Returns the information of the system state which is supported to read as like cpu, memory, networ
+        /// </summary>
+        /// <param name="packageName">The package name of the application</param>
+        /// <param name="dataType">The type of system state which wants to read. It should be one of the supported performance data types.</param>
```
mykola-mokhnach, line 11: Please enumerate the supported data types explicitly. Perhaps, it would make sense to have a separate enum for these

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// <param name="dataReadTimeout">The number of attempts to read (optional)</param>
```
mykola-mokhnach, line 12: what is the default read timeout?

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
```diff
@@ -0,0 +1,40 @@
...
+        ///                      [1478095200000, null, 1306300, 405997, 509, 46359, 370, 0, 3600000]]
+        ///        in case of network info :
+        ///            [[st, activeTime, rb, rp, tb, tp, op, bucketDuration],
+        ///                       [1478088000, null, null, 32115296, 34291, 2956805, 25705, 0, 3600],
+        ///                      [1478091600, null, null, 2714683, 11821, 1420564, 12650, 0, 3600],
+        ///                      [1478095200, null, null, 10079213, 19962, 2487705, 20015, 0, 3600],
+        ///                      [1478098800, null, null, 4444433, 10227, 1430356, 10493, 0, 3600]]
+        ///        in case of cpu info : [[user, kernel], [0.9, 1.3]]
+        /// </returns>
+        IList<object> GetSupportPerformanceData(string packageName, string dataType, int dataReadTimeout = 0);
```
mykola-mokhnach, line 33: does zero value mean the timeout is not limited?

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
```diff
@@ -0,0 +1,40 @@
...
+        
+        /// <summary>
+        /// Returns the information types of the system state which is supported to read as like cpu, memory, network traffic, and battery
```
mykola-mokhnach, line 36: It might be useful to provide an example of the output

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
mykola-mokhnach, line 11: it makes sense to mention these data types are returned by `GetSupportedPerformanceDataTypes()` call

src/Appium.Net/Appium/Android/AndroidDriver.cs
akinsolb: Ah, this is supposed to stated supported, but even that term is confusing, in order for someone to know what is supported they'd have to call GetPerformanceDataTypes.. thinking of making the datatype parameter an enum instead of a string

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
akinsolb: I can't even remember the specification for the timeout. When I do, i'll add it to the xml documentation

src/Appium.Net/Appium/Android/AndroidDriver.cs
mykola-mokhnach: I would say `GetPerformanceDataTypes` would be enough

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
mykola-mokhnach: actually, these data types are static: https://github.com/appium/appium-android-driver/blob/c1a8ee54b69e14940d4c2ecb766be7f75235aba2/lib/commands/performance.js#L12

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
mykola-mokhnach: actually, I've jus checked the source code. The argument name is confusing. The value means how many times the perf data retrieval operation should be retried in case of failure: https://github.com/appium/appium-android-driver/blob/c1a8ee54b69e14940d4c2ecb766be7f75235aba2/lib/commands/performance.js#L69

The default value is 2 with 1000ms delays between each retry

akinsolb: If I set this to 0 the appium logs doesn't display a second attempt, it just returns null.
1. I'm going to change the parameter to dataReadAttempts and update the param notes to make this clearer. 
2. Set the default value of dataReadAttempts to 1 because setting it to 0 returns null from the server (1.18).
3. Prevent users from setting 0 as the default value

akinsolb:
> @akinsolb Please provide the actual server error, perhaps I could fix it

https://gist.github.com/akinsolb/78d3343b39baf2f989f8e5a21c8c4a52 This is a run on an emulator (Android 10)

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
akinsolb: The summary is an example of an output

src/Appium.Net/Appium/Android/AndroidDriver.cs
akinsolb: done

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
akinsolb: made the above changes

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
akinsolb: done

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
mykola-mokhnach: I would rather create an overloaded method without this parameter. As far as I can see it is optional, so you could just skip sending it to the server: https://github.com/appium/appium-base-driver/blob/eba4635dcd8fe81899c3349d7e474acfb66037e7/lib/protocol/routes.js#L380

src/Appium.Net/Appium/Android/AndroidDriver.cs
```diff
@@ -213,7 +213,24 @@ public ConnectionType ConnectionType
 
         protected override RemoteWebElementFactory CreateElementFactory() => new AndroidElementFactory(this);
 
-        #region locking
+        #region Device Performance Data
+
+        public IList<object> GetPerformanceData(string packageName, string performanceDataType,
+            int dataReadAttempts = 1)
+        {
+            if (dataReadAttempts < 1) throw new ArgumentException($"{nameof(dataReadAttempts)} must be greater than 0");
```
mykola-mokhnach, line 221: 👍

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
akinsolb: I could either create an overload or perform a check on the driver method code, as it's optional, it already works as an overload to users

mykola-mokhnach:
I've pushed https://github.com/appium/appium-android-driver/pull/659 to fix the issue https://github.com/appium/appium-dotnet-driver/pull/422#issuecomment-687620106

You could check it out locally and verify now or wait until it's approved and is merged to beta

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
mykola-mokhnach: is it? We are always sending the `dataReadTimeout` arg in https://github.com/appium/appium-dotnet-driver/pull/422/files#diff-db1821172278290200c0e11b4a1a7b1dR120

akinsolb:
> I've pushed [appium/appium-android-driver#659](https://github.com/appium/appium-android-driver/pull/659) to fix the issue [#422 (comment)](https://github.com/appium/appium-dotnet-driver/pull/422#issuecomment-687620106)
> 
> You could check it out locally and verify now or wait until it's approved and is merged to beta

Thanks. I'll wait till it's merged into beta.

src/Appium.Net/Appium/Interfaces/IHasPerformanceData.cs
akinsolb: changes made

akinsolb: An alternative is making the dataReadAttempts argument nullable on the driver like so:

` public IList<object> GetPerformanceData(string packageName, string performanceDataType,
            int? dataReadAttempts = null) `
and checking for the value in the android command executor like so:

`public static object[] GetPerformanceData(IExecuteMethod executeMethod, string packageName, string dataType, int? dataReadTimeout)`

      `{
            var requiredKeys = new [] {"packageName", "dataType"};
            var requiredKeyValues = new object[] {packageName, dataType};
            if (dataReadTimeout is null)
                return executeMethod.Execute(AppiumDriverCommand.GetPerformanceData,
                    PrepareArguments(requiredKeys, requiredKeyValues)).Value as object[];
            requiredKeys.SetValue("dataReadTimeout", 2);
            requiredKeyValues.SetValue(dataReadTimeout, 2);
            return executeMethod.Execute(AppiumDriverCommand.GetPerformanceData,
                PrepareArguments(requiredKeys, requiredKeyValues)).Value as object[];
        }`

mykola-mokhnach (approved):
the cpuinfo fix has been published

PR #449, opened 2021-01-14
build: reenable GeneratePackageOnBuild

akinsolb:
## Change list

Project file

## Details

Reenabled the GeneratePackageOnBuild property because the release github action looks for a generated package for it to publish.
The file version is still removed so the assembly and package should not show an outdated version


akinsolb:
@mykola-mokhnach This reverts a change I made in the last PR. Unable to test this before merging unfortunately

mykola-mokhnach (approved):
I think it would make sense to leave a comment why this should not be changed

PR #468, opened 2021-10-07
refactor!: Update to work with Selenium 4

TroyWalshProf:
## Change list

Transition off Remote elements and drivers
Stop using platform specific WebElements and ElementFactories
Stop  using capabilities, instead leverage the base DriverOptions functionality
Transition from AddAdditionalCapability to AddAdditionalOption
Add 'App' as a known capability
Tag driver options so they are W3C compliant  
Transition from CommandInfo to HttpCommandInfo
 
## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Documentation
## Integration tests

## Details

Update to work with Selenium 4/to be W3C compliant


src/Appium.Net/Appium/AppiumOptions.cs
```diff
@@ -8,38 +11,160 @@ namespace OpenQA.Selenium.Appium
...
-        /// The dictionary of capabilities
+        /// Initializes a new instance of the <see cref="AppiumOptions"/> class.
         /// </summary>
-        private readonly AppiumCapabilities capabilities = new AppiumCapabilities();
+        public AppiumOptions() : base()
+        {
+            this.AddKnownCapabilityName(AppiumOptions.AutomationNameOption, "AutomationName property");
+            this.AddKnownCapabilityName(AppiumOptions.DeviceNameOption, "DeviceName property");
+            this.AddKnownCapabilityName(AppiumOptions.AppOption, "Application property");
+            this.AddKnownCapabilityName("app", "Application property");
```
mykola-mokhnach, line 30: why should we add it without the prefix? It is against the spec

src/Appium.Net/Appium/AppiumOptions.cs
```diff
@@ -8,38 +11,160 @@ namespace OpenQA.Selenium.Appium
...
-        /// Add new capabilities
+        /// Gets the vendor prefix to apply to Chromium-specific capability names.
         /// </summary>
-        /// <param name="capabilityName">Capability name</param>
-        /// <param name="capabilityValue">Capabilities value, which cannot be null or empty</param>
-        public override void AddAdditionalCapability(string capabilityName, object capabilityValue)
+        protected string VendorPrefix
         {
-            if (string.IsNullOrEmpty(capabilityName))
+            get { return "appium"; }
```
mykola-mokhnach, line 39: please move to a constant

src/Appium.Net/Appium/Service/AppiumCommandExecutor.cs
```diff
@@ -12,22 +12,28 @@
...
     internal class AppiumCommandExecutor : ICommandExecutor
     {
         private readonly AppiumLocalService Service;
-        private readonly Uri URL;
+        private static Uri URL;
+        private static TimeSpan CommandTimeout;
+
         private ICommandExecutor RealExecutor;
         private bool isDisposed;
-        private const string IdempotencyHeader = "X-Idempotency-Key";
```
mykola-mokhnach, line 27: What was the reason behind removing this feature?

test/integration/Android/ElementTest.cs
```diff
@@ -110,7 +112,7 @@ public void SetImmediateValueTest()
             var editElement =
                 _driver.FindElementByAndroidUIAutomator("resourceId(\"io.appium.android.apis:id/edit\")");
 
-            editElement.SetImmediateValue(value);
+            ///// TODO: Implement - editElement.SetImmediateValue(value);
```
mykola-mokhnach, line 115: is it a TODO for another PR?

src/Appium.Net/Appium/Service/AppiumCommandExecutor.cs
TroyWalshProf: Selenium 4 no longer allows us to directly interaction with HttpClient
See:
https://github.com/SeleniumHQ/selenium/blob/trunk/dotnet/src/webdriver/Remote/HttpCommandExecutor.cs

Would love to hear of something (there than reflection) that would work.

src/Appium.Net/Appium/AppiumOptions.cs
```diff
@@ -8,38 +11,160 @@ namespace OpenQA.Selenium.Appium
...
+            this.ValidateCapabilityName(name);
+            this.additionalAppiumOptions[name] = optionValue;
         }
 
         /// <summary>
+        /// Add new capabilities
+        /// </summary>
+        /// <param name="capabilityName">Capability name</param>
+        /// <param name="capabilityValue">Capabilities value, which cannot be null or empty</param>
+        [Obsolete("Use the temporary AddAdditionalOption method or the AddAdditionalChromeOption method for adding additional options")]
```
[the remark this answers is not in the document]
jimevans: Because the natural, “correct” name of the method _should_ be `AddAdditionalCapability`, to be consistent with other language bindings, the spec, documentation, and user expectations set over years of use. However, it wouldn’t be fair to users to remove an overload of the method that is in wide use nor to blindly change behavior of the method immediately with 4.0. I _want_the method to be `AddAdditionalCapability`, and expect to return to that syntax once there has been enough notice to users that the behavior has changed.

src/Appium.Net/Appium/Service/AppiumCommandExecutor.cs
jimevans: You can still do this with v4. The move to `HttpClient` broke it, because the `HttpWebRequest` object was no longer used, but we put the ability to modify headers back in [this commit](https://github.com/SeleniumHQ/selenium/commit/c66b1ea6bce7e09219512254b9f0b6bc31acf477)

src/Appium.Net/Appium/AppiumOptions.cs
TroyWalshProf: A bit torn on if we should have it or not.
If a user does something like this:
`appCapabilities.AddAdditionalOption("app", "Microsoft.WindowsCalculator_8wekyb3d8bbwe!App");`
*This currently just works with the latest Appium.net release
The setup will fail and tell them they never defined and application.
This way you get a useful error message.

@akinsolb - Would love to get your 2 cents here

src/Appium.Net/Appium/AppiumOptions.cs
```diff
@@ -8,38 +11,160 @@ namespace OpenQA.Selenium.Appium
...
+                name = optionName;
+            }
+            else
             {
-                throw new ArgumentException("Capability name may not be null an empty string.", "capabilityName");
+                name = $"{this.VendorPrefix}:{optionName}";
             }
 
-            this.capabilities[capabilityName] = capabilityValue;
+            this.ValidateCapabilityName(name);
```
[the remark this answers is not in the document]
TroyWalshProf: Nope, dropped the first one

src/Appium.Net/Appium/Service/AppiumCommandExecutor.cs
TroyWalshProf: Not sure of the best way to do this.
This looks like the easiest implementation
```
    internal class AppiumCommandExecutor : HttpCommandExecutor
    {
        private readonly AppiumLocalService Service;
        private bool isServiceDisposed;

        /// <summary>
        /// Initializes a new instance of the <see cref="AppiumCommandExecutor"/> class
        /// </summary>
        /// <param name="addressOfRemoteServer">Address of the WebDriver Server</param>
        /// <param name="timeout">The timeout within which the server must respond.</param>
        public AppiumCommandExecutor(Uri addressOfRemoteServer, TimeSpan timeout)
            : base(addressOfRemoteServer, timeout, true)
        {
            Service = null;
        }

        /// <summary>
        /// Initializes a new instance of the <see cref="AppiumCommandExecutor"/> class
        /// </summary>
        /// <param name="service">Appium local service</param>
        /// <param name="timeForTheServerResponding">The timeout within which the server must respond.</param>
        internal AppiumCommandExecutor(AppiumLocalService service, TimeSpan timeForTheServerResponding)
            : base(service.ServiceUrl, timeForTheServerResponding, true)
        {
            Service = service;
        }

        protected override void OnSendingRemoteHttpRequest(SendingRemoteHttpRequestEventArgs eventArgs)
        {
            eventArgs.AddHeader("X-Idempotency-Key", Guid.NewGuid().ToString());
            base.OnSendingRemoteHttpRequest(eventArgs);
        }

        protected override void Dispose(bool disposing)
        {
            if (!isServiceDisposed)
            {
                if (disposing)
                {
                    Service?.Dispose();
                }

                isServiceDisposed = true;
            }

            base.Dispose(disposing);
        }
    }
```
and it seems to work, but really terribly confident

src/Appium.Net/Appium/AppiumOptions.cs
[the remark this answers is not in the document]
mykola-mokhnach: cannot we just have both methods then and document the one as preferable? I don't really like to put a "time bomb" there, which will explode as soon as this temporary method would be deprecated.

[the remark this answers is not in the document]
jimevans: @mykola-mokhnach Appium, of course, is free to sculpt its API however it desires. Selenium, however, will be deprecating and renaming the methods at a future time. What will happen is this:
1. At some point, the `Obsolete` tag will be removed from `AddAdditionalCapability`, and added to `AddAdditionalOption`.
2. The `AddAdditionalOption` method will be removed from the Selenium language bindings.
The timeframe for each of these steps will be several minor releases during the 4.x release cycle.
The problem with `AddAdditionalCapability` today is that there are two overloads:
* `AddAdditionalCapability(string name, object value)`, defined in the base `DriverOptions` class, which adds the capability to the browser-specific capabilities (`goog:chromeOptions`, `moz:firefoxOptions`, etc.), and
* `AddAdditionalCapability(string name, object value, bool isGlobalCapability)`, defined in each browser-specific options class, which lets the user tell the bindings to add it to the top-level capabilities.

The boolean value logic is backward from what it ought to be to be put on the base class, which is the reason for this whole dance in the first place, to avoid breaking existing code, rather than silently changing the behavior of `AddAdditionalCapability` under the covers.

src/Appium.Net/Appium/Android/AndroidDriver.cs
```diff
@@ -25,11 +25,11 @@
 
 namespace OpenQA.Selenium.Appium.Android
 {
-    public class AndroidDriver<W> : AppiumDriver<W>, IFindByAndroidUIAutomator<W>, IFindByAndroidDataMatcher<W>,
+    public class AndroidDriver : AppiumDriver, IFindByAndroidUIAutomator<IWebElement>, IFindByAndroidDataMatcher<IWebElement>,
         IStartsActivity,
         IHasNetworkConnection, INetworkActions, IHasClipboard, IHasPerformanceData,
         ISendsKeyEvents,
-        IPushesFiles, IHasSettings where W : IWebElement
```
Dor-bl, line 32: we really want to do that? what is the reason behind it?

src/Appium.Net/Appium/AppiumOptions.cs
akinsolb: @TroyWalshProf only just seeing this. Think it's best to add the prefix. The driver options can override it.

PR #471, opened 2021-10-17
WIP: feat: add WindowsOption

akinsolb:
## Types of changes

## Documentation
## Integration tests

## Details
Add option class for windows driver
Updated AppiumOptions to match v5 changes without breaking changes


src/Appium.Net/Appium/AppiumOptions.cs
```diff
@@ -34,12 +92,60 @@ public override void AddAdditionalCapability(string capabilityName, object capab
...
+            }
+
+            return capabilities;
+        }
+
+        protected virtual Dictionary<string, object> BuildAppiumKnownOptionsDictionary()
+        {
+            var knownOptions = new Dictionary<string, object>();
+
+            if (!string.IsNullOrEmpty(App))
```
mykola-mokhnach, line 109: can we avoid some code duplication here?

src/Appium.Net/Appium/Windows/WindowsOptions.cs
```diff
@@ -0,0 +1,185 @@
...
+        public string PostRunScript { get; set; }
+
+
+        [Obsolete(
+            "Use the temporary AddAdditionalOption method or the browser-specific method for adding additional options")]
+        public override void AddAdditionalCapability(string capabilityName, object capabilityValue)
+        {
+            if (IsNullOrEmpty(capabilityName))
+            {
+                throw new ArgumentException("Capability name may not be null an empty string.", "capabilityName");
```
mykola-mokhnach, line 112: may not be not or an empty string

src/Appium.Net/Appium/Windows/WindowsOptions.cs
```diff
@@ -0,0 +1,185 @@
...
+            }
+
+            AddAdditionalOption(capabilityName, capabilityValue);
+        }
+
+        public override ICapabilities ToCapabilities()
+        {
+            var capabilities = GenerateDesiredCapabilities(false);
+
+            if (!IsNullOrEmpty(PlatformName))
```
mykola-mokhnach, line 122: can we avoid code duplication here?

src/Appium.Net/Appium/Windows/WindowsOptions.cs
```diff
@@ -0,0 +1,185 @@
...
+        private const string ExperimentalWebDriverOption = "ms:experimental-webdriver";
+        private const string SystemPortOption = "systemPort";
+        private const string PreRunOption = "prerun";
+        private const string PostRunOption = "postrun";
+
+        public WindowsOptions()
+        {
+            PlatformName = "windows";
+            AutomationName = "windows";
+            AddKnownCapabilityName(AutomationNameOption, "AutomationName property");
```
mykola-mokhnach, line 26: would it be necessary to update this list every time we add a new cap to the driver or users could do that on their own as well?

src/Appium.Net/Appium/Windows/WindowsOptions.cs
akinsolb: Yeah I'm going to call the base method, I just realised not all caps were being picked up because of that

src/Appium.Net/Appium/Windows/WindowsOptions.cs
akinsolb: It would be necessary to update this list every time we add a new cap. The base.addAdditionalOption() is advised for not yet added type safe options

src/Appium.Net/Appium/Windows/WindowsOptions.cs
```diff
@@ -0,0 +1,185 @@
...
+    {
+        private new string VendorPrefix = "ms";
+        private const string PlatformNameOption = "platformName";
+        private new const string AutomationNameOption = "automationName";
+        private new const string AppOption = "app";
+        private const string AppArgumentsOption = "appArguments";
+        private const string AppTopLevelWindowOption = "appTopLevelWindow";
+        private const string AppWorkingDirOption = "appWorkingDir";
+        private const string CreateSessionTimeoutOption = "createSessionTimeout";
+        private const string WaitForAppLaunchOption = "ms:waitForAppLaunch";
```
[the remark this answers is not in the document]
mykola-mokhnach: I mean that I can observe some unnecessary duplication where we define a separate constant for `ms` above and then still use the string itself in other constants instead of using the variable.

src/Appium.Net/Appium/AppiumOptions.cs
```diff
@@ -8,24 +9,81 @@ namespace OpenQA.Selenium.Appium
     /// </summary>
     public class AppiumOptions : DriverOptions
     {
+        private const string VendorPrefix = "appium";
+        private const string AutomationNameOption = "appium:automationName";
```
mykola-mokhnach, line 13: Can we use VendorPrefix constant instead of hardcoding `appium` prefix in all constants?

src/Appium.Net/Appium/Windows/WindowsOptions.cs
akinsolb: That goes against the purpose of Options classes in Selenium .Net, type safe options.

src/Appium.Net/Appium/AppiumOptions.cs
akinsolb: I'd imagine, derived classes can override it by supplying their own type safe options, but in this case, windows driver isn't w3c compliant

src/Appium.Net/Appium/Windows/WindowsOptions.cs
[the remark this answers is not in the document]
akinsolb: Oh I got rid of that constant.

src/Appium.Net/Appium/AppiumOptions.cs
mykola-mokhnach: so you mean it is impossible in dotnet to change `private const string AutomationNameOption = "appium:automationName"` to `rivate const string AutomationNameOption = $"{VendorPrefix}:automationName"` ?

akinsolb: Oh no, it isn't. Not yet anyway, string interpolation for const is in the upcoming language update I think. But I meant that derived option classes will provide their own type safe options with their vendor prefix set. This AppiumOptions class is a copy of what is in v5 branch, and is subject to multiple changes, especially defining which caps are appium, and stripping out what isn't. For the sake of these final v4 releases, it has no meaning

src/Appium.Net/Appium/Windows/WindowsOptions.cs
```diff
@@ -0,0 +1,174 @@
...
+            }
+
+            AddAdditionalOption(capabilityName, capabilityValue);
+        }
+
+        public override ICapabilities ToCapabilities()
+        {
+            var capabilities = base.ToCapabilities() as DesiredCapabilities;
+
+            if (!IsNullOrEmpty(AppArguments))
```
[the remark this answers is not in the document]
akinsolb: This is no different to a user setting caps. If a loop is to be involved, there will still need to be an if statement on the property, before adding it to a dictionary, then adding another method to loop that dictionary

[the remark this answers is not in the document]
mykola-mokhnach: I'd say suhc syntax might be shorter than the current one (pseudocode):

```
stringOptions = {
  AppArgumentsOption: AppArguments,
  AppTopLevelWindowOption: AppTopLevelWindow,
  ...
};
for (const key, value in stringOptions.items()) {
   if (!IsNullOrEmpty(value)) {
    capabilities.SetCapability(key, value);
   }
}
```

src/Appium.Net/Appium/Windows/WindowsOptions.cs
akinsolb: WindowsOption.AddAdditionalOption(string optionName, string optionValue) See [here](https://github.com/SeleniumHQ/selenium/blob/ee3ecbda341528d23152bcc1ca9e590c4e60c10c/dotnet/src/webdriver/DriverOptions.cs#L225)

PR #475, opened 2021-10-17
chore: Update MobileCapabilityType to support W3C protocol

Dor-bl:
## Change list

-Add prefix  'appium:' to required Mobile Capability Types
 
## Types of changes

## Documentation
## Integration tests

## Details

Update MobileCapabilityType to support W3C protocol.
**Will update documentation after 5.0 is released, for now there is no point.

### DIFF
--- a/src/Appium.Net/Appium/Enums/MobileCapabilityType.cs
+++ b/src/Appium.Net/Appium/Enums/MobileCapabilityType.cs
@@ -19,7 +19,7 @@ namespace OpenQA.Selenium.Appium.Enums
         /// <summary>
         /// Capability name used for the apllication setting.
         /// </summary>
-        public static readonly string App = "app";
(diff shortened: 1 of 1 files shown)


Dor-bl:
> Check #471

I don't see my changes there as well, so currently if a user will pass one of the mobile caps types it will fail, unless we want the user to pass a string as an option.

mykola-mokhnach:
Maybe I have just skipped something. @akinsolb might have better overview on the project

akinsolb:
@Dor-bl This change has been implicitly done in AppiumOptions class in the v5 branch. It takes any cap that doesn't have a prefix and adds the appium prefix. Ideally, these capability classes can be deleted, providing driver specific options add their own prefix. Given the comments on #471 I'm not sure if 'app' is an appium capability or a driver specific capability, especially in the windows instance.

PR #529, opened 2022-10-14
feat: Migrate AppiumLocalService to Appium 2

Dor-bl:
## Change list

-Removed the usage of /wd/hub which is no longer in use starting from appium2.0 server
-Minor code improvements at AppiumLocalService.cs
-Removal of redundant 'using' in test/integration/Properties/Resources.Designer.cs
 
## Types of changes

## Documentation
## Integration tests

## Details

### DIFF
--- a/src/Appium.Net/Appium/Service/AppiumLocalService.cs
+++ b/src/Appium.Net/Appium/Service/AppiumLocalService.cs
@@ -18,6 +18,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
+using System.Linq;
 using System.Net;
 using System.Runtime.CompilerServices;
 
@@ -40,11 +41,11 @@ namespace OpenQA.Selenium.Appium.Service
         public static AppiumLocalService BuildDefaultService() => new AppiumServiceBuilder().Build();
 
         internal AppiumLocalService(
-            FileInfo nodeJS, 
-            string arguments, 
-            IPAddress ip, 
+            FileInfo nodeJS,
+            string arguments,
+            IPAddress ip,
             int port,
-            TimeSpan initializationTimeout, 
+            TimeSpan initializationTimeout,
             IDictionary<string, string> environmentForProcess)
         {
             NodeJS = nodeJS;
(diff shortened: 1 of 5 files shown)


src/Appium.Net/Appium/Service/AppiumLocalService.cs
```diff
@@ -58,7 +58,7 @@ public class AppiumLocalService : ICommandServer
         /// <summary>
         /// The base URL for the managed appium server.
         /// </summary>
-        public Uri ServiceUrl => new Uri($"http://{IP.ToString()}:{Convert.ToString(Port)}/wd/hub");
+        public Uri ServiceUrl => new Uri($"http://{IP}:{Convert.ToString(Port)}");
```
mykola-mokhnach, line 61: this will make the service fully incompatible with Appium 1. Also, one could provide --base-path/-bp argument to change this path. We should be able to support such approach. At least Python and Java clients do

src/Appium.Net/Appium/Service/AppiumLocalService.cs
```diff
@@ -184,17 +184,17 @@ private bool Ping(TimeSpan span)
             Uri service = ServiceUrl;
             if (service.IsLoopback || IP.ToString().Equals(AppiumServiceConstants.DefaultLocalIPAddress))
             {
-                status = new Uri("http://localhost:" + Convert.ToString(Port) + "/wd/hub/status");
+                status = new Uri("http://localhost:" + Convert.ToString(Port) + "/status");
```
mykola-mokhnach, line 187: `localhost` might be pointing to an ipv6 address by default, so the status check won't work if the server is listening on ipv4 one. I would rather be quite careful while hardcoding it

src/Appium.Net/Appium/Service/AppiumLocalService.cs
Dor-bl: @mykola-mokhnach , already discussed the fact it will be incompatible with appium 1 with @akinsolb, this is the reason we will merge to release 5.0 branch.
I will try to look what I can do with the --base-path/-bp argument in order to support it fully.

src/Appium.Net/Appium/Service/AppiumLocalService.cs
Dor-bl: I assume this is something that wasn't handled by now anyway, or it's something new introduced with appium 2?

src/Appium.Net/Appium/Service/AppiumLocalService.cs
akinsolb: I didn't know the other bindings supported v1. It would be ideal if we could support both, where users can build their own client

src/Appium.Net/Appium/Service/AppiumLocalService.cs
mykola-mokhnach: it's unrelated to the server. The issue is a pure client one.
I do not know how it was implemented/working before, just pointing to a potential issue

src/Appium.Net/Appium/Service/AppiumLocalService.cs
Dor-bl: @akinsolb /@mykola-mokhnach Do you prefer I add the base path support as part of the arguments in the AppiumLocalService or it's better to create a dedicated var for it? just like we have IP and Port

mykola-mokhnach: I don't have any strong pref there. Do it the way which is closer/more straightforward to dotnet client users. In java and python service implementations we  fetch base path value from the corresponding CLI argument. `/` is used by default.

Dor-bl: I've pushed some changes to this PR so we will support base-path, but I still need to work on the ipv4 status issue.

src/Appium.Net/Appium/Service/AppiumLocalService.cs
Dor-bl: Not sure I understand the potential issue. 
When I send a request with the below request which includes the status check, the response returns OK. 
And the HostNameType is IPv4 with localhost.
image

image

akinsolb (approved):

PR #583, opened 2023-01-09
build: create XML documentation file on build and copy it to both binary output directories

### LINKED ISSUE #441: missing XML documentation in nuget package
Dear team,

the file `Appium.Net.xml` is missing in the nuget package _Applium.Net_.

Without this file, you can't see the documentation of the methods, etc. inside of Visual Studio.
This makes programming more tricky. Especially for Appium newcomers.

It would be nice if you add the XML documentation file to the nuget package.
To build that file, you just need to set one checkbox in the build settings of _Release_.

By the way:
In the old nuget package name _appium-dotnet-driver_ there was an xml file available.

Thanks,
Andreas

akinsolb:
## What types of changes are you proposing/introducing to the .NET client?

## Details
Create XML documentation file on build and copy it to both binary output directories. Currently a file is created in the project  root folder instead of the binary output directory.

closes #441


src/Appium.Net/Appium.Net.csproj
```diff
@@ -26,15 +26,28 @@
...
+    <LangVersion>8.0</LangVersion>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Debug|netstandard2.0|AnyCPU'">
+    <NoWarn>1701;1702;1591</NoWarn>
+  </PropertyGroup>
+
   <PropertyGroup Condition="'$(Configuration)|$(TargetFramework)|$(Platform)'=='Release|netstandard2.0|AnyCPU'">
-    <DocumentationFile>Appium.Net.xml</DocumentationFile>
+    <NoWarn>1701;1702;1591</NoWarn>
```
Dor-bl, line 42: @akinsolb can you explain what is the reason we need it?

akinsolb: The error code reduces the warnings on build. The xml file is used by consumers to see the documentation of the library. Currently this isn't included in the drop folder as we set the the file name. Removing it places it back in the default path, the build output directory.

Dor-bl (approved):

PR #629, opened 2023-08-10
refactor: Improve readability for Execute in AppiumCommandExecutor.cs

Dor-bl:
## Change list

1. Split the execute method into 3 small methods:
 a. HandleCommandException
 b. HandleCommandCompletion
 c. HandleNewSessionCommand

2. Since I've noticed flakiness when we send commands, I've implemented a retry mechanism based on response status in the Execute method. The method now attempts command execution up to MaxRetryAttempts times, checking the success status of the response before returning. I'm still in between about this change, not sure if it's a good approach at all, all I know is that it has improved the stability of the tests.

 
## Types of changes

## Documentation
## Integration tests

## Details


laolubenson:
@Dor-bl sorry i missed this PR. Did you find the reason why the execute method is flaky?

src/Appium.Net/Appium/Service/AppiumCommandExecutor.cs
```diff
@@ -136,12 +180,13 @@ private ICommandExecutor UpdateExecutor(Response result, ICommandExecutor curren
         /// <summary>
         /// Returns a new command executor if the response had directConnect.
         /// </summary>
-        /// <param name="result">The result of the command execution.</param>
+        /// <param name="response">The result of the command execution.</param>
```
Dor-bl, line 183: @KazuCocoam  I've fixed the wrong param name I noticed from your latest PR

Dor-bl:
@laolubenson, I'm in between here regarding the retry mechanism, it's not so elegant but does the trick when it comes to the flakiness of the execute method. 
Maybe @mykola-mokhnach  & @jlipps can add their input on why we have this issue in the first place that sometimes the server fails to succeed only after the 2nd try.

src/Appium.Net/Appium/Service/AppiumCommandExecutor.cs
```diff
@@ -25,6 +23,7 @@ internal class AppiumCommandExecutor : ICommandExecutor
         private ICommandExecutor RealExecutor;
         private bool isDisposed;
         private const string IdempotencyHeader = "X-Idempotency-Key";
+        private const int MaxRetryAttempts = 3;
```
mykola-mokhnach, line 26: do you know why command failures happen? We never retry session creation in other clients, especially implicitly
In java client , for example, the only situation where a failure could happen is the client read timeout (e.g. the session creation takes more than the client http timeout). Setting the read timeout to 10 minutes helped to solve the issue.

Dor-bl: Not sure why it happen.
So I'll remove the retry mechanism, anyway I didn't liked that approach.
I'll try playing with the read timeout if you suggest this solution. Can you provide a link to the place this magic happen on the Java client maybe?

Dor-bl: @mykola-mokhnach, to your question. 
here is an example on the first try I get noSuchDriver, but if I try again I receive a success status
first try:
image
2nd try:
image

mykola-mokhnach: do you also have the server log?

Dor-bl: sure. But I didn't see any logs related to `noSuchDriver` on the server.
https://gist.github.com/Dor-bl/a51a074d8e1aa4d8c02d242adb8067c8

mykola-mokhnach: I am not sure why it requests for log types on session creation, but NoSuchDriver error usually happens if the session dies because of some internal error/crash, so then all the following commands to the session identifier returned by NewSession API are not valid anymore.

Dor-bl: yes, I'm aware of this. but still, I get this error right after a new session creation and on the 2nd try it returns success

Dor-bl: @mykola-mokhnach, when you said to set client read timeout to 10 minutes, you mean here?
image

Dor-bl: same idea, no?
anyway, do you think it's more elegant to catch this case and retry again? only if noSuchdriver issue appear at first try

mykola-mokhnach: to me it feels like a workaround rather than a proper fix. 
I am not going to completely deny this PR, because I don't know much about the particular details of dotnet implementation and I trust you as the project maintainer.

Maybe you have anything to add on this @KazuCocoa ?

Dor-bl: So for now I'll leave the noSuchDriver issue. I made other required changes for the greater good. such as XML summaries and some code improvements

KazuCocoa: Sounds good in current pr.
I cannot recall properly, but probably I saw similar "no such driver" error message before. At that time was... maybe uia2/xcuitest's uia2 server/wda got failed or something in a new session request. Probably such a thing occurred...

but in this case, passing such an error to the caller (so test code) by default is probably good. Retrial could be an option though.

In my past experience, webdriverio had such retrial by default. it confused users so I needed to suggest disabling them first to make the failure reason clear. Thus maybe do not retry by default is basically better, but option could be. (this is my personal opinion)

KazuCocoa (approved):

src/Appium.Net/Appium/Service/AppiumCommandExecutor.cs
Dor-bl: Maybe I can add an option to to appium options class, that if the user set it's value to true it will activate retry for this case. this field will of course will be false by default.

PR #659, opened 2023-10-03
fix: Update Node path to compile with appium 2 and make him cross-platform

Dor-bl:
## List of changes

`appium.js` Windows node path used in the integration tests has changed on appium 2.
This PR addresses the the issue and makes sure it's generic for all users and compiles with appium 2.
In addition, I made a small refactor: Moved `Appium.js` path retrieval to `helpers.paths` class 

in theory, it should also fix one or more dotnet CI tests:
https://github.com/appium/dotnet-client/pull/650/checks

## Types of changes

## Documentation
## Integration tests

## Details

Please provide more details about changes if it is necessary. If there are new features you can provide code samples showing how they work and possible use cases. Also, you can create [gists](https://gist.github.com) with pasted C# code samples or put them here using markdown. 


test/integration/helpers/Paths.cs
```diff
@@ -0,0 +1,73 @@
...
+            else
+            {
+                // Handle unsupported platform
+                throw new PlatformNotSupportedException("Unsupported platform");
+            }
+
+            if (appiumJsPath.Contains("%USERPROFILE%"))
+            {
+                string userProfile = Environment.GetFolderPath(Environment.SpecialFolder.UserProfile);
+                userProfile = userProfile.Replace("\\", "/");
```
KazuCocoa, line 64: oh, i see

KazuCocoa (approved):

test/integration/helpers/Paths.cs
Dor-bl: Yea, basically it's not mandatory but feels more clean to have a proper path that way

test/integration/helpers/PathToWindowsNode
```diff
@@ -1 +1 @@
-C:/Program Files (x86)/Appium/node_modules/appium/bin/appium.js
\ No newline at end of file
+%USERPROFILE%/AppData/Roaming/npm/node_modules/appium/lib/appium.js
```
mykola-mokhnach, line 1: I believe this is not necessary and the Appium path could be detected dynamically based on the actual npm output:

win_path

Dor-bl: Hi @mykola-mokhnach, Not sure excalty what you mean by that, but we basically just use this path to test that we can pass a specific appium node path to the server.

mykola-mokhnach: I mean it is not necessary to hardcode the path to the appium script (which by the way should point to index.js in the module root rather than lib/appium.js)

Also the npm prefix itself should be retrieved dynamically from the npm config rather than hardcoded

Dor-bl: OK, I'll see what I can do about that.

Dor-bl: @mykola-mokhnach, same go to the Mac and Linux path?
we currently use:
/Applications/Appium.app/Contents/Resources/node_modules/appium/bin/appium.js

mykola-mokhnach: right, the algorithm should be cross-platform
I assume the only difference is that in *nix systems `npm` is an executable while on Windows it is a .cmd script

mykola-mokhnach: https://github.com/appium/java-client/blob/f875c3b86aa0f14e27a91d0a3d1f8e51d02e9054/src/main/java/io/appium/java_client/service/local/AppiumServiceBuilder.java#L147

Dor-bl: @mykola-mokhnach, I've pushed a big refactor, let me know if it's better now.
I will need to remove the other node's paths if we are good.
I also still need to test this on Unix/Mac system

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace Appium.Net.Integration.Tests.helpers
+{
+    internal class Npm
+    {
+        static private Platform envPlatform;
+
+        public static string GetNpmPrefixPath()
+        {
+            envPlatform = Platform.CurrentPlatform;
```
mykola-mokhnach, line 14: where envPlatform is used here?

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,70 @@
...
+            {
+                Console.WriteLine($"Error executing command: {ex.Message}");
+                return null;
+            }
+        }
+
+        private static string GetNpmExecutablePath()
+        {
+            string npmPath;
+            string command = envPlatform.IsPlatformType(PlatformType.Unix) || envPlatform.IsPlatformType(PlatformType.Mac) ? "which" : "where";
```
mykola-mokhnach, line 55: so `where.exe` is also able to find .cmd scripts?

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,70 @@
...
+            string[] lines = result?.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+            if (envPlatform.IsPlatformType(PlatformType.Windows))
+            {
+                npmPath = lines?.FirstOrDefault(line => line.EndsWith("npm.cmd"));
+            }
+            else 
+            {
+                npmPath = lines?.FirstOrDefault(line => line.EndsWith("npm"));
+            }
+            return npmPath;
```
mykola-mokhnach, line 67: should we throw a proper exception here if npm path cannot be detected?

test/integration/helpers/PathToWindowsNode
```diff
@@ -1 +1 @@
-C:/Program Files (x86)/Appium/node_modules/appium/bin/appium.js
```
mykola-mokhnach, line 1: does it still make sense to have two identical resources?

test/integration/helpers/Npm.cs
mykola-mokhnach, line 55: I would say it makes more sense to do it like `envPlatform.IsPlatformType(PlatformType.Windows) ? "where" : "which"`

test/integration/helpers/PathToWindowsNode
Dor-bl: I've marked them as obsolete, I'll remove them on the next commit

test/integration/helpers/Npm.cs
Dor-bl: We probably should 😉

test/integration/helpers/Npm.cs
Dor-bl: https://github.com/appium/dotnet-client/blob/0a1df1b57f4c3d3ce0d40632d08779b75a51e001/test/integration/helpers/Npm.cs#L55-L59

mykola-mokhnach: I would rather refactor that code because it creates an implicit contract/prependency between two methods., e.g. the one cannot be called unless the second has been invoked before.

Consider making envPlatform to a [lazy](https://stackoverflow.com/questions/4036812/c-sharp-lazy-loaded-automatic-properties) property instead

test/integration/helpers/PathToWindowsNode
Dor-bl: @mykola-mokhnach, while checking on MAC I encountered an issue with the path
index.js location on the Mac:
`/usr/local/lib/node_modules/appium/index.js`
image

but when running `which npm`, I get this npm prefix:
`/usr/local/bin/npm`
image

any suggestions on how to overcome this?

mykola-mokhnach: Looks like you are mixing npm install prefix and npm install path. The first one is where npm installs global modules and the second is just the path to the package manager binary itself

Dor-bl: So I should use a different command other than 'which npm' ?

mykola-mokhnach: Why not? Depending on what you want to achieve. If this is to find the npm script path then why not?

Dor-bl: so `npm list -g --depth=0` gives me the correct path, I just need to modify the `PathToNode` so it will work with Unix as well.
I will also need to see how it works with Windows when using the `npm list -g --depth=0`

mykola-mokhnach: usually the above command takes more time because it needs to scan multiple packages. Why not to use `npm config get prefix` or `npm -g root` ?

test/integration/helpers/Npm.cs
Dor-bl: I removed that part completely, after the latest commit.

test/integration/helpers/Paths.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Appium.Net.Integration.Tests.helpers;
+using Appium.Net.Integration.Tests.Properties;
+using System.IO;
+using System.Text;
+
+namespace Appium.Net.Integration.Tests.Helpers
+{
+    internal class Paths
+    {
+        private string _pathToCustomizedAppiumJs;
```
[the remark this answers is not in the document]
Dor-bl: this was the name since the beginning of time. If you have another suggestion, I'm all ears :)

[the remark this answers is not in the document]
mykola-mokhnach: _pathToAppiumPackageIndex

test/integration/helpers/Npm.cs
Dor-bl: done

test/integration/helpers/PathToWindowsNode
Dor-bl: removed

test/integration/helpers/Npm.cs
Dor-bl: @mykola-mokhnach, please review after my latest refactor and let me know if there is still a need for further refactor

test/integration/helpers/PathToWindowsNode
Dor-bl: Moved to `npm -g root`

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,70 @@
...
+                process.Start();
+                string output = process.StandardOutput.ReadToEnd();
+                process.WaitForExit();
+
+                return output;
+            }
+            catch (Exception ex)
+            {
+                Console.WriteLine($"Error executing command: {ex.Message}");
+                return null;
```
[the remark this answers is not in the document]
Dor-bl: So Should I remove the entire Try-catch in your opinion?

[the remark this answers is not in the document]
mykola-mokhnach: depending on what you want to achieve. 
This is the point of exceptions - they need to be thrown if there is an exceptional satiation in the workflow.

In this particular example I can propose three approaches:
- Let the code throw a generic exception
- Catch the exception and rethrow a more specific one, which describes the situation better for the end-user (e.g. instead of having ENOENT message show something like `Node Package Manager (npm) cannot be found. Make sure Node.js is installed and present in PATH`)
- Let the code throw a custom exception and catch it on a higher level to show a more specific error message to the user

In general I expect exceptions to be helpful and provide hints/workarounds on how to avoid them.

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,73 @@
...
+                        Arguments = arguments,
+                        RedirectStandardOutput = true,
+                        UseShellExecute = false,
+                        CreateNoWindow = true,
+                    }
+                };
+
+                process.Start();
+                string output = process.StandardOutput.ReadToEnd();
+                process.WaitForExit(timeoutMilliseconds);
```
mykola-mokhnach, line 45: does this call also verify the exit code?

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,73 @@
...
+{
+    internal class Npm
+    {
+
+        public static string GetNpmPrefixPath()
+        {
+
+            string npmPath = "npm";
+
+            if (Platform.CurrentPlatform.IsPlatformType(PlatformType.Windows))
```
mykola-mokhnach, line 16: why we only do it for windows?

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,73 @@
...
+            string result = RunCommand("where", "npm");
+            string npmPath;
+
+            string[] lines = result?.Split(new[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+
+            npmPath = lines?.FirstOrDefault(line => line.EndsWith("npm.cmd"));
+
+            if (string.IsNullOrWhiteSpace(npmPath))
+            {
+                throw new ApplicationException("NPM path not found.");
```
mykola-mokhnach, line 67: I believe this error message could be improved

test/integration/helpers/PathToNode
```diff
@@ -0,0 +1 @@
+appium/index.js
```
mykola-mokhnach, line 1: Not sure about the file name here. This is the path to appium package index, not to the node

test/integration/helpers/Paths.cs
```diff
@@ -0,0 +1,39 @@
...
+            {
+                if (_pathToAppiumPackageIndex == null)
+                {
+                    GetAppiumJsPath();
+                }
+                return _pathToAppiumPackageIndex;
+            }
+        }
+
+        private void GetAppiumJsPath()
```
mykola-mokhnach, line 24: GetAppiumPackageIndexPath

test/integration/helpers/Paths.cs
```diff
@@ -0,0 +1,39 @@
...
+        {
+            byte[] bytes;
```
mykola-mokhnach, line 26: not sure about conventions in C#, but I usually define vars close to places where they are actually used

test/integration/helpers/Npm.cs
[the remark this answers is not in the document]
Dor-bl: Since we are in RunCommand() method, a specific exception for NPM might be tricky since if in the future we would like to use this for other cases we can't throw the above error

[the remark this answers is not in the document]
Dor-bl: on the other end, we are in npm class :)

test/integration/helpers/PathToNode
Dor-bl: PathToPackageIndex?

test/integration/helpers/Paths.cs
```diff
@@ -0,0 +1,39 @@
...
+            string appiumJsPath;
+            string npmPath;
+           
+            bytes = Resources.PathToNode;
+            appiumJsPath = Encoding.UTF8.GetString(bytes); 
+
+            npmPath = Npm.GetNpmPrefixPath();
+
+            string tempPath = Path.Combine(npmPath, appiumJsPath);
+            _pathToAppiumPackageIndex = tempPath.Replace("\\", "/");
```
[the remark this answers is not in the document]
mykola-mokhnach: Then just keep an array of path components in the resource and join then dynamically in the code. The joiner function will ensure a proper separator

test/integration/helpers/Npm.cs
mykola-mokhnach: I get it. What I wanted to say is that `npm` might also not exist on other systems if it is not installed or not added to PATH

mykola-mokhnach: Which might result into a weird generic exception from process builder afterwards

test/integration/helpers/Npm.cs
Dor-bl: it does not, you see a point in adding this verification?

Dor-bl:
> Btw, it would be interesting to know why this code is only being used in `test` package. Is the service builder in the sources not working in the same way?

It works a bit differently there. 

https://github.com/appium/dotnet-client/blob/main/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs#L154-L212

test/integration/helpers/Npm.cs
Dor-bl: throw new NpmNotFoundException($"NPM executable not found at path: {npmPath}. Please make sure the NPM executable is installed and check the configured path.");

test/integration/helpers/Paths.cs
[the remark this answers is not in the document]
Dor-bl: Do you mean something like that?
appium
index.js

test/integration/helpers/Npm.cs
mykola-mokhnach: yes. Any non-zero response would mean a failure

test/integration/helpers/Paths.cs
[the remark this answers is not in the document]
mykola-mokhnach: yes

mykola-mokhnach:
> > Btw, it would be interesting to know why this code is only being used in `test` package. Is the service builder in the sources not working in the same way?
> 
> It works a bit differently there.
> 
> https://github.com/appium/dotnet-client/blob/main/src/Appium.Net/Appium/Service/AppiumServiceBuilder.cs#L154-L212

The idea is similar though. My point is we can just maximally reuse the existing stuff from there instead of reinventing a wheel.

Dor-bl:
@mykola-mokhnach, Unfortunately, I no longer have access to a MAC/Unix machine, so I think we should prioritize the remaining comments that are still open, I prefer not to refactor again without proper ENV to check for regression.
I know the current code works for sure

test/integration/ServerTests/AppiumLocalServerLaunchingTest.cs
```diff
@@ -41,25 +37,8 @@ public void BeforeAll()
...
-                return;
-            }
-            if (isLinux)
-            {
-                bytes = Resources.PathToLinuxNode;
-                _pathToCustomizedAppiumJs = Encoding.UTF8.GetString(bytes);
-                return;
-            }
+            Paths paths = new Paths();
+            _pathToAppiumPackageIndex = paths.PathToAppiumPackageIndex;
```
mykola-mokhnach, line 41: do we need to keep it as a property here if we get always get the path from the corresponding paths property?

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,86 @@
...
+                {
+                    process.Start();
+                    string output = process.StandardOutput.ReadToEnd();
+                    string errorOutput = process.StandardError.ReadToEnd();
+                    process.WaitForExit(timeoutMilliseconds);
+
+                    int exitCode = process.ExitCode;
+                    if (exitCode != 0)
+                    {
+                        throw new ApplicationException($"Command exited with code {exitCode}. Error: {errorOutput}");
```
mykola-mokhnach, line 43: we should also log which command

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,86 @@
...
+                    }
+
+                    return output;
+                }
+            }
+            catch (Exception)
```
mykola-mokhnach, line 49: I usually don't do that. If we want to map some particular exceptions to NpmNotFoundException then it would make sense to explicitly enumerate their classes. And eben in that case it is necessary to keep the original cause. "Muting" the generic exception class would make it hard to debug in the future

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,86 @@
...
+                npmPath = lines?.FirstOrDefault(line => !string.IsNullOrWhiteSpace(line) && line.EndsWith("npm.cmd"));
+            }
+            else
+            {
+                npmPath = lines?.FirstOrDefault(line => !string.IsNullOrWhiteSpace(line));
+            }
+
+            if (string.IsNullOrWhiteSpace(npmPath))
+            {
+                throw new NpmNotFoundException("NPM executable not found. Please make sure the NPM executable is installed and check the configured path.");
```
mykola-mokhnach, line 75: and check the configured PATH environment variable

mykola-mokhnach: also the default message would work here

test/integration/ServerTests/AppiumLocalServerLaunchingTest.cs
Dor-bl: Either we keep it as a property or I will need to create the Paths obj each time I need this property.

test/integration/helpers/Paths.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// Combines the path components from the appiumJsPath with the npmPath.
+        /// </summary>
+        /// <remarks>
+        /// This function reads a byte array from the PathToPackageIndex resource, converts it to a string,
+        /// processes the relative path to remove '\r\n', and combines it with the npm prefix path.
+        /// The resulting combined path is assigned to the _pathToAppiumPackageIndex variable.
+        /// </remarks>
+        private void GetAppiumPackageIndexPath()
+        {
+            byte[] bytes = Resources.PathToPackageIndex;
```
mykola-mokhnach, line 35: this seems a bit overcomplicated to me. Cannot we just keep it as a simple constant array?

test/integration/helpers/Paths.cs
```diff
@@ -0,0 +1,44 @@
...
+
+        /// <summary>
+        /// Combines the path components from the appiumJsPath with the npmPath.
+        /// </summary>
+        /// <remarks>
+        /// This function reads a byte array from the PathToPackageIndex resource, converts it to a string,
+        /// processes the relative path to remove '\r\n', and combines it with the npm prefix path.
+        /// The resulting combined path is assigned to the _pathToAppiumPackageIndex variable.
+        /// </remarks>
+        private void GetAppiumPackageIndexPath()
```
mykola-mokhnach, line 33: this method does not get anything as its return type is void. 
Consider renaming it to InitAppiumPackageIndexPath

test/integration/helpers/Npm.cs
Dor-bl: went for the first comment

test/integration/ServerTests/AppiumLocalServerLaunchingTest.cs
mykola-mokhnach: fine with me. I was just confused by the intermediate paths variable. Perhaps we could simply do _pathToAppiumPackageIndex = new Paths().PathToAppiumPackageIndex

test/integration/helpers/Npm.cs
Dor-bl: Something like that is more acceptable?

If have exitCode !=0 then we throw ApplicationException and then rethrow it after we catch it 

otherwise, we will throw NpmNotFoundException()
```
catch (ApplicationException ex)
{
    throw ex;
}

catch (Exception)
{
    throw new NpmNotFoundException();
}
```

test/integration/helpers/Paths.cs
Dor-bl: Just to drop that file and have an array? we can...

test/integration/helpers/Npm.cs
mykola-mokhnach: > catch (Exception)

This part looks potentially dangerous to me. Again, only specific exception classes should be enumerated in the catch clause

Dor-bl: FIxed. Hopefully :)

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                    else if (exitCode != 0)
+                    {
+                        throw new ApplicationException($"Command: `{command}` exited with code {exitCode}. Error: {errorOutput}");
+                    }
+
+                    return output;
+                }
+            }
+            catch (ApplicationException ex)
```
mykola-mokhnach, line 54: this syntax could be simplified: https://code-maze.com/csharp-catch-multiple-exceptions/

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,105 @@
...
+                {
+                    process.Start();
+                    string output = process.StandardOutput.ReadToEnd();
+                    string errorOutput = process.StandardError.ReadToEnd();
+                    _ = process.WaitForExit(timeoutMilliseconds);
+
+                    int exitCode = process.ExitCode;
+                    if (exitCode == 1)
+                    {
+                        throw new NpmUnknownCommandException($"Command: `{arguments}` exited with code {exitCode}. Error: {output}");
```
mykola-mokhnach, line 44: I think we may be more explicit about the exception message here if we know that exit code 1 means an invalid npm command

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,105 @@
...
+            catch (Win32Exception)
+            {
+                throw;
+            }
+            catch (NpmUnknownCommandException)
+            {
+                throw;
+            }
+            catch (Exception)
+            {
```
mykola-mokhnach, line 68: what I actually meant could be found at https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/exceptions/creating-and-throwing-exceptions

paragraph:

> When an argument to a method causes an exception. In this case, the original exception should be caught and an [ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception) instance should be created. The original exception should be passed to the constructor of the [ArgumentException](https://learn.microsoft.com/en-us/dotnet/api/system.argumentexception) as the [InnerException](https://learn.microsoft.com/en-us/dotnet/api/system.exception.innerexception) parameter

mykola-mokhnach: I believe the above article contains a decent tutorial on general exceptions design

test/integration/helpers/Npm.cs
Dor-bl: It's not always the case when exit code =1.
Could be other issues other than npm related

mykola-mokhnach: then what is the reason for making exit code 1 so special?

Dor-bl: I forgot to push the last commit :(

test/integration/helpers/Npm.cs
Dor-bl: I removed the ones I didn't have anything special to do there.

test/integration/helpers/Npm.cs
Dor-bl: I removed the ExitCode != 0 and kept only exitCode == 1 which is the one I saw while testing

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,91 @@
...
+                        throw new NpmUnknownCommandException($"Command: `{arguments}` exited with code {exitCode}. Error: {output}");
+                    }
+
+                    return output;
+                }
+            }
+            catch (Win32Exception ex) when (command.Contains("npm"))
+            {
+                Console.WriteLine(ex.Message);
+                throw new NpmNotFoundException($"mpm not found under {command}", ex);
```
mykola-mokhnach, line 56: typo npm

test/integration/helpers/Npm.cs
```diff
@@ -0,0 +1,91 @@
...
+                    if ((exitCode == 1) && command.Contains("npm"))
+                    {
+                        Console.WriteLine($"npm Error upon command: `{arguments}`. {output}");
+                        throw new NpmUnknownCommandException($"Command: `{arguments}` exited with code {exitCode}. Error: {output}");
+                    }
+
+                    return output;
+                }
+            }
+            catch (Win32Exception ex) when (command.Contains("npm"))
```
mykola-mokhnach, line 53: Should we enclose the whole method body into try/catch? I believe this exception is only thrown upon some particular API call. I usually try to minimise exception blocks to a very necessary

mykola-mokhnach (approved):

test/integration/helpers/Npm.cs
Dor-bl: thanks

test/integration/helpers/Npm.cs
Dor-bl: I can basically wrap the code up until `process.Start()`

PR #796, opened 2024-06-20
test: Fix ios ElementTests

Dor-bl:
## List of changes

- Remove the Deprecated  IosUiAutomation Test
- Add 2 new tests for ClassName 

## Types of changes

## Documentation
## Integration tests

## Details

Please provide more details about changes if it is necessary. If there are new features you can provide code samples showing how they work and possible use cases. Also, you can create [gists](https://gist.github.com) with pasted C# code samples or put them here using markdown. 

### DIFF
--- a/test/integration/IOS/ElementTest.cs
+++ b/test/integration/IOS/ElementTest.cs
@@ -56,7 +56,7 @@ namespace Appium.Net.Integration.Tests.IOS
         {
             try
             {
-                var switchElement = _driver.FindElement(MobileBy.AccessibilityId("XCUIElementTypeSwitch"));
+                var switchElement = _driver.FindElement(MobileBy.ClassName("XCUIElementTypeSwitch"));
             }   
             catch (NoSuchElementException ex)
             {
--- a/test/integration/IOS/ElementTest.cs
+++ b/test/integration/IOS/ElementTest.cs
@@ -54,8 +54,14 @@ namespace Appium.Net.Integration.Tests.IOS
         [Test]
         public void FindElementMobileByClassNameTest()
         {
-            var switchElement = _driver.FindElement(MobileBy.ClassName("XCUIElementTypeSwitch"));
-            Assert.That(switchElement, Is.Not.Null);
+            try
+            {
+                var switchElement = _driver.FindElement(MobileBy.AccessibilityId("XCUIElementTypeSwitch"));
+            }   
+            catch (NoSuchElementException ex)
+            {
+                Assert.Fail("Element not found, exception: " + ex.Message);
+            }
         }
 
     }
--- a/test/integration/IOS/ElementTest.cs
+++ b/test/integration/IOS/ElementTest.cs
@@ -35,7 +35,7 @@ namespace Appium.Net.Integration.Tests.IOS
             By byAccessibilityId = new ByAccessibilityId("ComputeSumButton");
             Assert.Multiple(() =>
             {
-                Assert.That(_driver.FindElement(MobileBy.ClassName("UIAWindow")).FindElement(byAccessibilityId).Text, Is.Not.Null);
+                Assert.That(_driver.FindElement(MobileBy.ClassName("UIAWindow")).FindElement(byAccessibilityId).Text, Is.EqualTo("Compute Sum"));
                 Assert.That(_driver.FindElement(MobileBy.ClassName("UIAWindow")).FindElements(byAccessibilityId), Is.Not.Empty);
             });
         }
--- a/test/integration/IOS/ElementTest.cs
+++ b/test/integration/IOS/ElementTest.cs
@@ -6,7 +6,7 @@ using OpenQA.Selenium.Appium.iOS;
 
 namespace Appium.Net.Integration.Tests.IOS
 {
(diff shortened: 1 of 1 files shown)


test/integration/IOS/ElementTest.cs
```diff
@@ -35,20 +35,23 @@ public void FindByAccessibilityIdTest()
             By byAccessibilityId = new ByAccessibilityId("ComputeSumButton");
             Assert.Multiple(() =>
             {
-                Assert.That(Is.Not.EqualTo(_driver.FindElements(MobileBy.ClassName("UIAWindow"))[1].FindElement(byAccessibilityId).Text), null);
-                Assert.That(_driver.FindElements(MobileBy.ClassName("UIAWindow"))[1].FindElements(byAccessibilityId), Is.Not.Empty);
+                Assert.That(_driver.FindElement(MobileBy.ClassName("UIAWindow")).FindElement(byAccessibilityId).Text, Is.Not.Null);
```
mykola-mokhnach, line 38: I assume findElement should throw an exception if nothing is found

Dor-bl: It does

test/integration/IOS/ElementTest.cs
```diff
@@ -35,20 +35,23 @@ public void FindByAccessibilityIdTest()
...
+        public void FindElementsByClassNameTest()
         {
-            By byIosUiAutomation = new ByIosUIAutomation(".elements().withName(\"Answer\")");
-            Assert.Multiple(() =>
-            {
-                Assert.That(_driver.FindElements(MobileBy.ClassName("UIAWindow"))[1].FindElement(byIosUiAutomation).Text, Is.Not.Null);
-                Assert.That(_driver.FindElements(MobileBy.ClassName("UIAWindow"))[1].FindElements(byIosUiAutomation), Is.Not.Empty);
-            });
+            By byClassName = new ByClassName("XCUIElementTypeTextField");
+            Assert.That(_driver.FindElements(byClassName), Has.Count.EqualTo(2));
```
mykola-mokhnach, line 47: I usually add timers for conditions like this, e.g. "verify the amount of found elements is equal or more than 2 within 5 seconds ". They work better in slow CI env and are more reliable in general.

test/integration/IOS/ElementTest.cs
```diff
@@ -35,20 +35,23 @@ public void FindByAccessibilityIdTest()
...
+        }
+
+        [Test]
+        public void FindElementMobileByClassNameTest()
+        {
+            var switchElement = _driver.FindElement(MobileBy.ClassName("XCUIElementTypeSwitch"));
+            Assert.That(switchElement, Is.Not.Null);
```
mykola-mokhnach, line 54: same as above

test/integration/IOS/ElementTest.cs
mykola-mokhnach: Then it would be never equal to null. The result is either an element or an exception thrown

Dor-bl: The Assert is done on the text of the element.
So if there is no text it will be empty, and test will fail

mykola-mokhnach: we may also make the check more strict and verify the text is equal to some predefined value. But the current check is also ok then.

test/integration/IOS/ElementTest.cs
Dor-bl: We are using `_driver.Manage().Timeouts().ImplicitWait = Env.ImplicitTimeoutSec;`
isn't that enough in this case?

test/integration/IOS/ElementTest.cs
Dor-bl: now we check the exact text value

test/integration/IOS/ElementTest.cs
Dor-bl: Do you mean the timer or the element not found exception?

test/integration/IOS/ElementTest.cs
mykola-mokhnach: The way implicit timeouts are applied only helps to wait for until any element exists. This means such wait will be finished if the given locator matches at least one element in the example above instead of two, which creates a potential race condition for apps where elements pop up on screen in serial order

test/integration/IOS/ElementTest.cs
mykola-mokhnach: yes, something like [this](https://stackoverflow.com/questions/9417213/how-do-i-check-no-exception-occurred-in-my-mstest-unit-test) would probably better reflect the purpose of this test

test/integration/IOS/ElementTest.cs
Dor-bl: I've harnessed the WebDriverWait usage for that cause. looks better?

test/integration/IOS/ElementTest.cs
Dor-bl: Fixed

mykola-mokhnach (approved):

KazuCocoa (approved):

PR #818, opened 2024-08-17
feat: Override selenium default command timeout

Dor-bl:
## List of changes

Override and increase the default HTTP requests command timeout from 1 minute to 10 minutes.
 
## Types of changes

- [x] New feature (non-breaking change that adds functionality or value)

## Documentation
## Integration tests
- [ ] Have you provided integration tests for your changes? (required for Bugfix or New feature)

## Details

### DIFF
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -38,7 +38,7 @@ namespace OpenQA.Selenium.Appium
         /// <summary>
         /// The default command timeout for HTTP requests in a AppiumDriver instance.
         /// </summary>
-        private static new readonly TimeSpan DefaultCommandTimeout = TimeSpan.FromSeconds(600.0);
+        protected static new readonly TimeSpan DefaultCommandTimeout = TimeSpan.FromSeconds(600.0);
 
         #region Constructors
 
--- a/src/Appium.Net/Appium/AppiumDriver.cs
+++ b/src/Appium.Net/Appium/AppiumDriver.cs
@@ -35,6 +35,11 @@ namespace OpenQA.Selenium.Appium
     {
         private const string NativeApp = "NATIVE_APP";
 
+        /// <summary>
+        /// The default command timeout for HTTP requests in a AppiumDriver instance.
+        /// </summary>
+        private static new readonly TimeSpan DefaultCommandTimeout = TimeSpan.FromSeconds(600.0);
+
         #region Constructors
 
         public AppiumDriver(ICommandExecutor commandExecutor, ICapabilities appiumOptions)


src/Appium.Net/Appium/AppiumDriver.cs
```diff
@@ -35,6 +35,11 @@ public abstract class AppiumDriver : WebDriver,
     {
         private const string NativeApp = "NATIVE_APP";
 
+        /// <summary>
+        /// The default command timeout for HTTP requests in a AppiumDriver instance.
+        /// </summary>
+        private static new readonly TimeSpan DefaultCommandTimeout = TimeSpan.FromSeconds(600.0);
```
mykola-mokhnach, line 41: where this value is used?

Dor-bl: https://github.com/appium/dotnet-client/blob/30d9dec6f4966b9bbd8ac687a2576efda8c0ac1c/src/Appium.Net/Appium/AppiumDriver.cs#L73

https://github.com/appium/dotnet-client/blob/30d9dec6f4966b9bbd8ac687a2576efda8c0ac1c/src/Appium.Net/Appium/AppiumDriver.cs#L78

https://github.com/appium/dotnet-client/blob/30d9dec6f4966b9bbd8ac687a2576efda8c0ac1c/src/Appium.Net/Appium/AppiumDriver.cs#L94

https://github.com/appium/dotnet-client/blob/30d9dec6f4966b9bbd8ac687a2576efda8c0ac1c/src/Appium.Net/Appium/AppiumDriver.cs#L99

asos-LaoluTairuAkinsola:
@Dor-bl ten minutes is crazy? Is that in-line with the other clients?

src/Appium.Net/Appium/AppiumDriver.cs
mykola-mokhnach: Do I understand it right, - is it the same property defined protected/public in the parent class?

mykola-mokhnach:
> @Dor-bl ten minutes seems crazy. Is that in-line with the other clients?

Yes it is. Sometimes the session startup for mobile session can take quite lot of time. For example, to boot an iOS simulator on a slow CI node.

mykola-mokhnach:
Also, the server and drivers usually limit command timings as well, for example via newCommandTimeout, proxy timeout, etc, which means killing the client connection due to read timeout must be used as a last resort. In most of cases such timeout would mean a bug or a missing timeout handler at the server side.

KazuCocoa:
I wondered if it would be nice to have a troubleshooting and add this time thing in our test code as an example rather than overriding the existing one in selenium to avoid unexpected modification in the parent (selenium) side. Just a personal preference.

src/Appium.Net/Appium/AppiumDriver.cs
Dor-bl: Correct

Dor-bl:
> I wondered if it would be nice to have a troubleshooting and add this time thing in our test code as an example rather than overriding the existing one in selenium to avoid unexpected modification in the parent (selenium) side. Just a personal preference.

You mean part of the Wiki? Because I don't think most users look in the test code.

KazuCocoa:
Yea, this is not strong opinion for the default value itself though.

When we override the value, probably it would be nice to address it in somewhere as well, docstring or wiki? in this dotnet client case. How to modify the timeout in Wiki also should help (as a general documentation thing)

Btw, the commit prefix could be `feat`. I saw some users had error handling for client side read timeout as well (real devices) to fail fast, so modifying the default value as a patch could change their behavior.

src/Appium.Net/Appium/AppiumDriver.cs
mykola-mokhnach: I am not an expert in dotnet, but in other languages it is usually not a good practice "reduce" member visibility level in descendants. That is why it makes sense to make this field either protected or public if it was defined as protected in the parent class.
Or simply rename it if you prefer to keep that private.

Dor-bl: Changed to protected

Dor-bl:
> Yea, this is not strong opinion for the default value itself though.
> 
> When we override the value, probably it would be nice to address it in somewhere as well, docstring or wiki? in this dotnet client case. How to modify the timeout in Wiki also should help (as a general documentation thing)
> 
> Btw, the commit prefix could be `feat`. I saw some users had error handling for client side read timeout as well (real devices) to fail fast, so modifying the default value as a patch could change their behavior.

@KazuCocoa, Let me know what you think.
@mykola-mokhnach, we need to decide if we stay with the Wiki for now and abandon this PR.

mykola-mokhnach (approved):

mykola-mokhnach:
> we need to decide if we stay with the Wiki for now and abandon this PR.

I don't want to abandon anything. The change makes sense and does fix potential user issues

KazuCocoa (approved):
Thank you!

PR #836, opened 2024-09-25
build(deps): Migration from `Newtonsoft.Json` to `System.Text.Json` package

Dor-bl:
## List of changes

- The selenium dotnet client recently moved to the `System.Text.Json` package. Why shouldn't we? :)
- Replace `Newtonsoft.Json` with `System.Text.Json`
- Added error handling for JSON deserialization and fallback logic for environment variables.

## Types of changes

- [x] New feature (non-breaking change that adds functionality or value)

## Documentation
## Integration tests
- [x] Have you provided integration tests for your changes? (required for Bugfix or New feature)

## Details

### DIFF
--- a/test/integration/helpers/Env.cs
+++ b/test/integration/helpers/Env.cs
@@ -15,29 +15,33 @@ namespace Appium.Net.Integration.Tests.helpers
 
         private static void Init()
         {
+            _env = new Dictionary<string, JsonElement>
+            {
+                { "DEV", JsonDocument.Parse("true").RootElement }, 
+                { "isRemoteAppiumServer", JsonDocument.Parse("false").RootElement }, 
+                { "remoteAppiumServerUri", JsonDocument.Parse("\"http://localhost:4723\"").RootElement }
+            };
+
+            if (_initialized) return;
+
             try
             {
-                if (!_initialized)
+                _initialized = true;
+                var path = AppDomain.CurrentDomain.BaseDirectory;
+                var sr = new StreamReader(path + "env.json");
+                var jsonString = sr.ReadToEnd();
+                _env = JsonSerializer.Deserialize<Dictionary<string, JsonElement>>(jsonString, new JsonSerializerOptions
                 {
-                    _initialized = true;
-                    var path = AppDomain.CurrentDomain.BaseDirectory;
-                    var sr = new StreamReader(path + "env.json");
-                    var jsonString = sr.ReadToEnd();
-                    _env = JsonSerializer.Deserialize<Dictionary<string, JsonElement>>(jsonString, new JsonSerializerOptions
-                    {
-                        PropertyNameCaseInsensitive = true
-                    });
-                }
+                    PropertyNameCaseInsensitive = true
+                });
             }
             catch (JsonException jsonEx)
             {
                 Console.WriteLine($"Error parsing JSON: {jsonEx.Message}");
-                _env = [];
             }
             catch (Exception ex)
             {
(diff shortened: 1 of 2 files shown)


test/integration/helpers/Env.cs
```diff
@@ -23,19 +23,28 @@ private static void Init()
...
+                    {
+                        PropertyNameCaseInsensitive = true
+                    });
                 }
             }
-            catch
+            catch (JsonException jsonEx)
             {
-                _env = new Dictionary<string, string>();
+                Console.WriteLine($"Error parsing JSON: {jsonEx.Message}");
```
mykola-mokhnach, line 34: is it ok to mute errors here?

test/integration/helpers/Env.cs
```diff
@@ -23,19 +23,28 @@ private static void Init()
...
             }
         }
 
-        private static bool IsTrue(string val)
+        private static bool IsTrue(object val)
         {
-            val = val?.ToLower().Trim();
-            return (val == "true") || (val == "1");
+            val = val?.ToString().ToLower().Trim();
+            return val is (object)"true" or (object)"1";
```
mykola-mokhnach, line 47: I would rather put the `or` part into parentheses to explicitly reflect the priority

test/integration/helpers/Env.cs
```diff
@@ -23,19 +23,28 @@ private static void Init()
...
+                _env = [];
             }
         }
 
-        private static bool IsTrue(string val)
+        private static bool IsTrue(object val)
         {
-            val = val?.ToLower().Trim();
-            return (val == "true") || (val == "1");
+            val = val?.ToString().ToLower().Trim();
```
mykola-mokhnach, line 46: would this work properly if val is null? In js I'd have something like `val?.ToString()?.ToLower()?.Trim()`

test/integration/helpers/Env.cs
mykola-mokhnach: AFAIK is may be used as a type-checking operator or for pattern matching. Strings comparison could be done using Equals

test/integration/helpers/Env.cs
Dor-bl: We were muting them before as well, so yeah. I just added this 2 cases with proper message

mykola-mokhnach: in such case you may assign the default value to _env before the `try` block to avoid unnecessary code duplication

mykola-mokhnach: generally, exceptions mute is considered a suboptimal coding practice. Check, for example, https://xp123.com/code-smells-in-exceptions/

test/integration/helpers/Env.cs
Dor-bl: VS thinks other wise in regard to the parentheses
image

Dor-bl: Strings comparison is now done using Equals

test/integration/helpers/Env.cs
Dor-bl: added default values to _env before the try.
I agree that silencing expectations is not ideal, but in this case, which is just internal testing for contributors, there is no need to fail with an exception.

test/integration/helpers/Env.cs
Dor-bl: since I added the default values for _env, we shouldn't worry of a null case

KazuCocoa (approved):

mykola-mokhnach (approved):

PR #874, opened 2024-11-29
BREAKING CHANGE: `AndroidDriver` to use modern `mobile:` commands for Lock, IsLocked and Unlock

Dor-bl:
## List of changes

- Replaced deprecated JSONWP commands with `mobile: lock`, `mobile: isLocked`, and `mobile: unlock` extensions.
- Updated the `Lock` method to accept an optional `seconds` parameter for auto-unlock functionality.
- Updated the `Unlock` method to support required `key` and `type` parameters, with optional `strategy` and `timeoutMs` arguments.
- Improved code robustness with null checks and parameter validation.
 
## Types of changes

- [ ] New feature (non-breaking change that adds functionality or value)
- [x] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [x] **Test fix** (non-breaking change that improves test stability or correctness)

## Documentation
- [ ] Have you proposed a file change/ PR with Appium to update documentation? 
## Integration tests
- [x] Have you provided integration tests for your changes? (required for Bugfix, New feature, or Test fix)

## Details

### DIFF
--- a/src/Appium.Net/Appium/Android/AndroidDriver.cs
+++ b/src/Appium.Net/Appium/Android/AndroidDriver.cs
@@ -287,21 +287,67 @@ namespace OpenQA.Selenium.Appium.Android
 
         #region Device Interactions
 
-        /**
-        * This method locks a device.
-        */
-        public void Lock() => AppiumCommandExecutionHelper.Lock(this, 0);
+        /// <summary>
+        /// Locks the device. Optionally, unlocks it after a specified number of seconds.
+        /// </summary>
+        /// <param name="seconds">
+        /// The number of seconds after which the device will be automatically unlocked. 
+        /// Set to 0 or leave it empty to require manual unlock.
+        /// </param>
+        /// <exception cref="WebDriverException">Thrown if the command execution fails.</exception>
+        public void Lock(int? seconds = null)
+        {
+            var parameters = new Dictionary<string, object>();
+
+            if (seconds.HasValue && seconds.Value > 0)
+            {
+                parameters["seconds"] = seconds.Value;
+            }
+
+            ExecuteScript("mobile: lock", parameters);
+        }
 
         /// <summary>
         /// Check if the device is locked
         /// </summary>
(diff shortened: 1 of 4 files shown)


src/Appium.Net/Appium/Android/AndroidDriver.cs
```diff
@@ -287,21 +287,67 @@ public IList<string> GetPerformanceDataTypes() =>
...
+        public void Lock(int? seconds = null)
+        {
+            var parameters = new Dictionary<string, object>();
+
+            if (seconds.HasValue && seconds.Value > 0)
+            {
+                parameters["seconds"] = seconds.Value;
+            }
+
+            ExecuteScript("mobile: lock", parameters);
```
KazuCocoa, line 307: You could add try/except and fallback to the legacy one like https://github.com/appium/python-client/blob/e4f06abf1f6bc302bde373071d1e3c007a67c03a/appium/webdriver/extensions/hw_actions.py#L40-L47 does. Then, maybe the change doesn't need to be a "breaking" one since the existing one will do the same

Dor-bl: The "breaking" is more for the `unlock` method that now requires additional parameters

Dor-bl: I also see there is a a todo on that fallback
https://github.com/appium/python-client/blob/e4f06abf1f6bc302bde373071d1e3c007a67c03a/appium%2Fwebdriver%2Fextensions%2Fhw_actions.py#L45
So isn't it redundant?

KazuCocoa: Yea, the todo is correct. We haven't decided when we completely remove them, though. I'm good either, just I brought up the idea here

KazuCocoa (approved):
lgtm.

Perhaps it would be nice to include similar breaking changes in one release if we have others

Dor-bl:
I'll try to find some time, sure

PR #1098, opened 2026-07-26
docs: drop appiumpro.com references

### LINKED ISSUE #22499: Remove appiumpro.com related links from our repos
Because the domain is no longer related to us.

alaahong:
## Summary
Removes `appiumpro.com` references from this repository, since the domain is no longer maintained by the Appium team (see appium/appium#22499).

Follows the same approach as appium/appium#22500 — links are removed rather than replaced, unless removal would break surrounding prose, in which case the URL is inlined as plain text.

Closes appium/appium#22499

## Checklist
- [x] I have read the [Contributing Guide](https://appium.io/docs/en/latest/contributing/)
- [x] I have signed the CLA
- [x] Lint and unit tests pass locally with my changes
- [x] No documentation update needed (only stale links removed)

### DIFF
--- a/src/Appium.Net/Appium/Service/AppiumClientConfig.cs
+++ b/src/Appium.Net/Appium/Service/AppiumClientConfig.cs
@@ -29,7 +29,7 @@ namespace OpenQA.Selenium.Appium.Service
         /// <summary>
         /// Gets or sets the directConnect feature availability.
         /// If this flag is true and the target server supports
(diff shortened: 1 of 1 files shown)


KazuCocoa:
We need to have alternative documentation for them instead of simple drop for individual pages

alaahong:
> We need to have alternative documentation for them instead of simple drop for individual pages

Get the point. Caused that's the legacy link, seems no need to create new doc file into Appium docs? just replace with new link for same content reference.

KazuCocoa (approved):

PR #1121, opened 2026-09-14
ci: stabilize iOS simulator startup before functional tests

KazuCocoa:
The iOS job currently relies on `simulator-action` boot completion and leaves Simulator UI startup to the first Appium session. In [run 34810153703](https://github.com/appium/dotnet-client/actions/runs/34810153703), the first iOS fixture failed because the simulator did not finish booting within 120 seconds, while subsequent fixtures ran.

Align startup with the existing [WebDriverAgent workflow](https://github.com/appium/WebDriverAgent/blob/3e8aa7de81f254dbb0876baa9e9173c16b55b3a0/.github/workflows/functional-test.yml) and [XCUITest driver workflow](https://github.com/appium/appium-xcuitest-driver/blob/65585c013009dd3f3dd4e85cb89f7dca185d017c/.github/workflows/functional-test.yml):

- Open the selected Xcode's Simulator UI before preparing the simulator, and list devices/runtimes for diagnostics.
- Adapt the driver's dependency-free [idle helper](https://github.com/appium/appium-xcuitest-driver/blob/65585c013009dd3f3dd4e85cb89f7dca185d017c/scripts/ci/wait-for-simulator-idle.mjs) to wait for background services immediately before functional tests. Keep the upstream defaults: three consecutive samples below 20% aggregate child-process CPU, sampled every two seconds, with a 150-second best-effort limit. A three-minute step timeout bounds unexpected hangs.
- Add dependency-free CLI tests and run them in the iOS job.

Validation: all five helper tests pass (selected-device isolation, aggregate CPU and consecutive-sample reset, timeout, disappearing simulator, missing UDID); Node syntax check, actionlint, and `git diff --check` pass. The complete iOS suite remains for PR CI. The observed timeout motivates this alignment, but does not establish CPU contention as its root cause; improved flake rate needs repeated hosted-runner results. The helper preserves upstream warning-and-continue behavior when the simulator disappears or does not settle.


Dor-bl (changes requested):
The iOS-tests job completed in 47m 49s, compared to historical runs of ~22–25 minutes. Approximately ~25 minutes of unnecessary overhead was introduced in this run.

See my comment for more details.

Dor-bl:
@KazuCocoa Can you please check the merge conflicts?

PR #1129, opened 2026-09-17
test: add iOS ShakeDevice and PerformTouchID coverage

Dor-bl:
## PR title

test: add iOS ShakeDevice and PerformTouchID coverage

## Related issue

Closes # n/a

## List of changes

- Add `test/integration/IOS/DeviceInteractionTest.cs` covering `IOSDriver.ShakeDevice()` and `IOSDriver.PerformTouchID(bool)`, neither of which had any existing test coverage.
- `PerformTouchIdTest` first enrolls Touch ID on the simulator via `mobile: enrollBiometric` (required before `sendBiometricMatch` has anything to match against), then exercises both the match and no-match paths.

## Types of changes

- [x] **New test coverage** (non-breaking change that adds tests for existing, previously untested functionality)

## Tests

- [x] Integration tests

**How they run:** Picked up by the existing `FullyQualifiedName~IOS` CI filter (tagged `[Category("iOS")]`), no CI changes needed.

## Documentation
- [x] Not applicable (no user-facing behaviour change, e.g. tests, CI or maintenance only)

## Details

Verified locally against a real booted iOS 26.0 "iPhone 17" simulator with a local Appium server + XCUITest driver: both tests pass. Note for anyone running this locally without CI's `LOCAL_PREBUILT_WDA` env var set — a cold WebDriverAgent build can take long enough that the client-side HTTP request drops before the server responds (looks like a connection failure, isn't). Grab a prebuilt WDA first: `appium driver run xcuitest download-wda-sim --platform=ios --outdir=<dir>` and set `LOCAL_PREBUILT_WDA` to the extracted `.app` path.


KazuCocoa (approved):
Did it work? I mean not the response wise, but actually if shake worked

Dor-bl:
Checked — typed text into a field, called `ShakeDevice()`, and iOS's built-in "Undo Typing?" alert appeared. So it's a real shake motion event, not just a no-op API call. The test itself only asserts `DoesNotThrow` though — happy to strengthen it to assert on that alert if useful.

Dor-bl:
> Did it work? I mean not the response wise, but actually if shake worked

